PR #45, opened 2018-03-31
Raise error if trying to train before adding any examples.

darrengarvey:
The error only appears in the JS console but at least now the error message
is clear. Before, there would be an error when trying to use the null value
`controllerDataset.xs.shape[0]`.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/45)

### DIFF
--- a/webcam-transfer-learning/index.js
+++ b/webcam-transfer-learning/index.js
@@ -35,6 +35,9 @@ let model;
 const controllerDataset = new ControllerDataset(NUM_CLASSES);
 
 async function train() {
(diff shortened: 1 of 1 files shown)


nsthorat:
:lgtm_strong:

Thanks for the error message!

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/tensorflow/tfjs-examples/45#-:-L8yCM7N8_An4nOIuYDt:bcvc6o2)*

PR #53, opened 2018-04-04
Fixing stddev computation ?

jeromeetienne:
It seems that this -1 may not be suitable. Im not sureThis is acording to my low level in stats. but a check on the internet seems to agree. [link on kahnacademy](https://www.khanacademy.org/math/probability/data-distributions-a1/summarizing-spread-distributions/a/calculating-standard-deviation-step-by-step)

 Im not even sure. i havent even tested the code. 

ps: thanks for the work on tensorflow.js/deeplearnjs, i having a lot of fun learning it :)

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/53)

### DIFF
--- a/polynomial-regression/index.js
+++ b/polynomial-regression/index.js
@@ -88,7 +88,7 @@ function stddev(vector) {
   for (const x of vector) {
     squareSum += (x - vectorMean) * (x - vectorMean);
   }
-  return Math.sqrt(squareSum / (vector.length - 1));
+  return Math.sqrt(squareSum / vector.length);
 }
 
 // Normalize a vector by its mean and standard deviation.


jeromeetienne:
i got trouble signing the cla (i got multiple google account and the cla page doesnt allow to login on the proper account.

on the other hand this is a super small modification, so i guess you can retype it. thus no legal issue :) thanks again for your work :)

jeromeetienne:
i fix i understood my mistake, the goal is to do a stddev on a sample, not a population.

so the formula is correct and my fix is not. 

sorry for the inconvenience, closing now

PR #59, opened 2018-04-11
Fix vega-lite config to fix console warnings

bileschi:
The vega config had quantitative style for both 'x' and 'y', which leads to a warning in vega-lite.  The expectation is that one axis will be of type 'temporal' or (in this case) 'ordinal'.

You can play around with vega-lite configs easily with 
https://vega.github.io/editor/#/edited

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/59)

### DIFF
--- a/polynomial-regression-core/ui.js
+++ b/polynomial-regression-core/ui.js
@@ -58,14 +58,14 @@
        {
          'mark': 'point',
          'encoding': {
-           'x': {'field': 'x', 'type': 'quantitative'},
+           'x': {'field': 'x', 'type': 'ordinal'},
            'y': {'field': 'y', 'type': 'quantitative'}
          }
        },
        {
          'mark': 'line',
          'encoding': {
-           'x': {'field': 'x', 'type': 'quantitative'},
+           'x': {'field': 'x', 'type': 'ordinal'},
            'y': {'field': 'pred', 'type': 'quantitative'},
            'color': {'value': 'tomato'}
          },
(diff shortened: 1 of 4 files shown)


nsthorat:
:lgtm_strong:

---

Review status: 0 of 4 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/tensorflow/tfjs-examples/59#-:-L9qlk6j7PAt93_pQ8CP:b-jbsf8z)*

polynomial-regression-core/ui.js
```diff
@@ -58,14 +58,14 @@
        {
          'mark': 'point',
          'encoding': {
-           'x': {'field': 'x', 'type': 'quantitative'},
+           'x': {'field': 'x', 'type': 'ordinal'},
```
SkyWox: This is quantitative data, not ordinal data. Changing the x axis to ordinal results in a poorly rendered line as seen below: 

image

bileschi:
Thanks SkyWox.  Do you know how to correctly address this warning from Vega, in the event that both axes are quantitative?
```
[Warning] Cannot clearly determine orientation for "line" since both x and y channel encode continuous fields. In this case, we use vertical by default
```

nsthorat:
Stan, can you send a PR to bring back quantitative for polynomial-regression? I couldn't find an easy way to silence the warning. It looks really broken with ordinal though so let's undo that (just for the one example that breaks).

PR #70, opened 2018-05-07
Fix build error in windows when set NODE_ENV

OuYancey:
Hey! Why there is no ISSUE tab in this repo?
1. Build Error in windows. There are two ways to fix this:
  - add cross-env lib
  - only remove `NODE_ENV=production`, we may not need this?

2. Not smooth line in [online example](https://storage.googleapis.com/tfjs-examples/polynomial-regression-core/dist/index.html). 
see the capture:(CHROME V66)
not_smooth
u'd better update the online example.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/70)

### DIFF
--- a/polynomial-regression-core/package.json
+++ b/polynomial-regression-core/package.json
@@ -14,14 +14,15 @@
     "vega-lite": "^2.3.1"
   },
   "scripts": {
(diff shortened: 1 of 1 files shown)


OuYancey:
emm, I just found all examples has `NODE_ENV=production`...

PR #72, opened 2018-05-07
Add accuracy off of test data.

nkreeger:
Here is eval off of test data for class percentage. I also have one tweak to only return a pitch card for each pitch type. We can drop this if we don't like the cards.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/72)

### DIFF
--- a/baseball-node/src/abstract-pitch-model.ts
+++ b/baseball-node/src/abstract-pitch-model.ts
@@ -16,9 +16,8 @@
  */
 
 import * as tf from '@tensorflow/tfjs';
+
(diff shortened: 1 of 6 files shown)


dsmilkov:
:lgtm_strong:

---

Reviewed 4 of 6 files at r1.
Review status: 3 of 6 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/72#-:-LBwAEot-ZQqw_YKqk_N:b-jbsf8z)*

PR #84, opened 2018-05-11
 Pac-man demo: predictions must be manually disposed

davidsoergel:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/84)


nsthorat:
:lgtm_strong:

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---

*[webcam-transfer-learning/index.js, line 150 at r1](https://reviewable.io/reviews/tensorflow/tfjs-examples/84#-LCFvEun6bI1R22llM0n:-LCFvEun6bI1R22llM0o:bkujmi) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/43116fa98eca4d3285989a147e257fad925cb706/webcam-transfer-learning/index.js#L150)):*
> ```JavaScript
> 
>     const classId = (await predictedClass.data())[0];
>     predictedClass.dispose();
> ```

this isn't actually a memory leak because .data() releases the texture, but I think I agree that's sort of an implementation detail. LGTM.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/84#-:-LCFvL8CAXG1_arGJqsm:b-jbsf8z)*

davidsoergel:
Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[webcam-transfer-learning/index.js, line 150 at r1](https://reviewable.io/reviews/tensorflow/tfjs-examples/84#-LCFvEun6bI1R22llM0n:-LCFwfCEBR9hXG5ISOWm:b3z0ihd) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/43116fa98eca4d3285989a147e257fad925cb706/webcam-transfer-learning/index.js#L150)):*
<details><summary><i>Previously, nsthorat (Nikhil Thorat) wrote…</i></summary><blockquote>

this isn't actually a memory leak because .data() releases the texture, but I think I agree that's sort of an implementation detail. LGTM.
</blockquote></details>

Wait, really?  The docs don't mention that, and a quick code spelunking revealed no disposal within data() (though of course I may have just overlooked it or misunderstood something).

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/84)*

nsthorat:
Yeah, so it doesn't dispose the Tensor but it releases the underlying texture (which is why we have to do memory management in the first place). If this Tensor goes out of scope, regular GC will kick in (and no memory leak). If you need that tensor again, we reupload it to a new texture.

https://github.com/tensorflow/tfjs-core/blob/master/src/kernels/backend_webgl.ts#L1046

PR #103, opened 2018-06-25
Add example: LSTM text generation in the browser

caisq:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/103)


caisq:
Example screenshot:
image

nsthorat:
Let me know if you have questions about the comments I left, sorry that I'm kind of requesting a large refactor. I think this PR is super awesome and promising, nice work getting it to work!!

---

Review status: 0 of 1 LGTMs obtained

---

*[lstm-text-generation/index.js, line 38 at r3](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LFsrnZl71mYsnNwV_FL:-LFsrnZl71mYsnNwV_FM:b9ukl4b) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/6530029a52a19b79a787df3e96762974fe5afb40/lstm-text-generation/index.js#L38)):*
> ```JavaScript
> 
> // UI controls.
> const testText = document.getElementById('test-text');
> ```

I would move all of the ui related stuff to a ui.js file. All the ML should live in this file (and data processing in a data.js file). When a user comes to this example, they should try to see things that are as high-level as possible w.r.t non-ML stuff, and get immediately to the meat of the ML. See how this index.js file is structured:
https://github.com/tensorflow/tfjs-examples/blob/master/webcam-transfer-learning/index.js

If I'm a new person this file is pretty intimidating and I can't find the relevant bits for me

---

*[lstm-text-generation/nn-text-generator.js, line 30 at r3](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LFsnbg74IsMdYZZRqG8:-LFsnbg74IsMdYZZRqG9:b6him3a) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/6530029a52a19b79a787df3e96762974fe5afb40/lstm-text-generation/nn-text-generator.js#L30)):*
> ```JavaScript
>  * @returns {Array} Shuffled array.
>  */
> function shuffle(array) {
> ```

we have a util.shuffle: https://github.com/tensorflow/tfjs-core/blob/master/src/util.ts#L23

---

*[lstm-text-generation/nn-text-generator.js, line 55 at r3](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LFsniR03rv_7ImCdHBj:-LFsniR03rv_7ImCdHBk:b-rwlb63) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/6530029a52a19b79a787df3e96762974fe5afb40/lstm-text-generation/nn-text-generator.js#L55)):*
> ```JavaScript
>  * @returns {number} A zero-based sample index.
>  */
> function sampleOneFromMultinomial(probs) {
> ```

tf.multinomial?

---

*[lstm-text-generation/nn-text-generator.js, line 78 at r3](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LFsnpePA7IZgBneaM2L:-LFsnpePA7IZgBneaM2M:b9nf6mj) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/6530029a52a19b79a787df3e96762974fe5afb40/lstm-text-generation/nn-text-generator.js#L78)):*
> ```JavaScript
>  * - Generating random text based on the LSTM model.
>  */
> export class NeuralNetworkTextGenerator {
> ```

So the way we've been trying to structure these examples is to put the meat of the ML inside of the index.js file. When people look to see how to make their own generative text model they'll eventually end up here after finding the call site in index.js. The idea would be for them to easily transplant the example, understanding how the pieces of it works, however this file seems like it's generally useful as a neural network text generator utility (capturing corner cases, etc). I wonder if it's better to simply have much "stupider" code that is much easier to parse. As it is, even I'm having a hard time following some of this.

Few comments inline where I think it could be simplified:

---

*[lstm-text-generation/nn-text-generator.js, line 122 at r3](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LFsp6Bu5lRt9GOSMdsB:-LFsp6Bu5lRt9GOSMdsC:b2pluff) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/6530029a52a19b79a787df3e96762974fe5afb40/lstm-text-generation/nn-text-generator.js#L122)):*
> ```JavaScript
>    * @returns {number} Length of training text data.
>    */
>   textLen() {
> ```

if you un-class-ify this, you dont need all the code for these getters

---

*[lstm-text-generation/nn-text-generator.js, line 154 at r3](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LFsqZu9EAHtchv2GTwy:-LFsqZu9EAHtchv2GTwz:b-41k74o) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/6530029a52a19b79a787df3e96762974fe5afb40/lstm-text-generation/nn-text-generator.js#L154)):*
> ```JavaScript
>    *   `ys` has the shape of `[numExamples, this.charSetSize]`.
>    */
>   nextDataEpoch(numExamples) {
> ```

this could be in a data.js file

---

*[lstm-text-generation/nn-text-generator.js, line 206 at r3](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LFsqfZC4lO6_INve1Z8:-LFsqfZC4lO6_INve1Z9:b-uf6obs) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/6530029a52a19b79a787df3e96762974fe5afb40/lstm-text-generation/nn-text-generator.js#L206)):*
> ```JavaScript
> 
>       if (characterCallback != null) {
>         await characterCallback(winnerChar);
> ```

you can make this simpler if this goes in index.js and calls an explicit "ui.update()" or something like that -- the genericness of this callback will confuse people

---

*[lstm-text-generation/nn-text-generator.js, line 220 at r3](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LFsqnBw5hq1cNVI59hg:-LFsqnBw5hq1cNVI59hh:b-fxmwhj) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/6530029a52a19b79a787df3e96762974fe5afb40/lstm-text-generation/nn-text-generator.js#L220)):*
> ```JavaScript
> 
>   /**
>    * Create LSTM model.
> ```

This method should be one of the first thing people see inside of index.js when they come to this example

---

*[lstm-text-generation/nn-text-generator.js, line 230 at r3](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LFsqwRu07A755YQJbwl:-LFsqwRu07A755YQJbwm:b-4zgszb) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/6530029a52a19b79a787df3e96762974fe5afb40/lstm-text-generation/nn-text-generator.js#L230)):*
> ```JavaScript
>     let createFromScratch;
>     if (loadModelIfAvailable === true) {
>       const modelsInfo = await tf.io.listModels();
> ```

I don't think this example should have any of the logic for loading and saving models, it just complicates it. Think of examples as just explaining a single concept. Once you remove all this logic for saving, this becomes much much simpler and easier to grok

---

*[lstm-text-generation/nn-text-generator.js, line 259 at r3](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LFsr6SlEYW00IUhQfDS:-LFsr6SlEYW00IUhQfDT:b-9xfmen) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/6530029a52a19b79a787df3e96762974fe5afb40/lstm-text-generation/nn-text-generator.js#L259)):*
> ```JavaScript
>         const lstmLayerSize = lstmLayerSizes[i];
>         if (!(lstmLayerSize > 0)) {
>           throw new Error(
> ```

this error throwing should be at the UI level, not complicating this method

---

*[lstm-text-generation/nn-text-generator.js, line 270 at r3](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LFsrDQN1vLlH77BWzRD:-LFsrDQN1vLlH77BWzRE:b-sb4rso) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/6530029a52a19b79a787df3e96762974fe5afb40/lstm-text-generation/nn-text-generator.js#L270)):*
> ```JavaScript
>       }
>       this.model.add(tf.layers.dense({
>         units: this._charSetSize,
> ```

when you move this, it will become something like ui.getCharSetSize(), just like we do here: https://github.com/tensorflow/tfjs-examples/blob/master/webcam-transfer-learning/index.js#L80

---

*[lstm-text-generation/nn-text-generator.js, line 282 at r3](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LFsrPUT-7fbO7gBXrUF:-LFsrPUU-jVByWK6mGJH:b-xkxsbs) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/6530029a52a19b79a787df3e96762974fe5afb40/lstm-text-generation/nn-text-generator.js#L282)):*
> ```JavaScript
>    */
>   compileModel(learningRate) {
>     const optimizer = tf.train.rmsprop(learningRate);
> ```

compile should live alongside creating the model

---

*[lstm-text-generation/nn-text-generator.js, line 291 at r3](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LFsrRl_3Jl05tmPdgTc:-LFsrRl_3Jl05tmPdgTd:b-aw2wdw) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/6530029a52a19b79a787df3e96762974fe5afb40/lstm-text-generation/nn-text-generator.js#L291)):*
> ```JavaScript
>    * Remove the locally saved model from IndexedDB.
>    */
>   async removeModel() {
> ```

these methods should go away (remove and save), same with checking status

---

*[lstm-text-generation/nn-text-generator.js, line 362 at r3](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LFsrWJu8v98MUEKmFfv:-LFsrWJu8v98MUEKmFfw:bw5phu7) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/6530029a52a19b79a787df3e96762974fe5afb40/lstm-text-generation/nn-text-generator.js#L362)):*
> ```JavaScript
>           },
>           onBatchEnd: async (batch, logs) => {
>             embed(
> ```

ui.batchEnd() or something along those lines

---

*[lstm-text-generation/nn-text-generator.js, line 384 at r3](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LFsr_plDaZWZx6ihfZs:-LFsr_plDaZWZx6ihfZt:b-wt1z3d) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/6530029a52a19b79a787df3e96762974fe5afb40/lstm-text-generation/nn-text-generator.js#L384)):*
> ```JavaScript
>           },
>           onEpochEnd: async (epochs, log) => {
>             console.log(
> ```

Update the UI instead of the console?

---

*[lstm-text-generation/nn-text-generator.js, line 423 at r3](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LFsrhkiEn9fiWUL4ncB:-LFsrhkiEn9fiWUL4ncC:b-mfu8sh) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/6530029a52a19b79a787df3e96762974fe5afb40/lstm-text-generation/nn-text-generator.js#L423)):*
> ```JavaScript
>    * Get the set of unique characters from text.
>    */
>   _getCharSet() {
> ```

this could live in a data.js file, same with all the other data-related stuff

---

*[lstm-text-generation/package.json, line 12 at r3](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LFrvspT3Ar1dHSy36sE:-LFrvspUCcY-FF-4Ng7G:bf6y0rc) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/6530029a52a19b79a787df3e96762974fe5afb40/lstm-text-generation/package.json#L12)):*
> ```JSON
>   },
>   "dependencies": {
>     "@tensorflow/tfjs": "^0.11.7",
> ```

use ~ here and below

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-:-LFst3aE-GbulFFnidn-:b-h36i0c)*

caisq:
Thanks for the very helpful review, Nikhil. I've addressed your comments. See replies below.

In addition, I made a change to the UI: Now there is a new "Seed text" input box, which allows the user to modify the seed text used during text generation.

---

Review status: 0 of 1 LGTMs obtained

---

*[lstm-text-generation/index.js, line 38 at r3](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LFsrnZl71mYsnNwV_FL:-LFyB-QSEv1q1dSanGwF:bgygb0r) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/6530029a52a19b79a787df3e96762974fe5afb40/lstm-text-generation/index.js#L38)):*
<details><summary><i>Previously, nsthorat (Nikhil Thorat) wrote…</i></summary><blockquote>

I would move all of the ui related stuff to a ui.js file. All the ML should live in this file (and data processing in a data.js file). When a user comes to this example, they should try to see things that are as high-level as possible w.r.t non-ML stuff, and get immediately to the meat of the ML. See how this index.js file is structured:
https://github.com/tensorflow/tfjs-examples/blob/master/webcam-transfer-learning/index.js

If I'm a new person this file is pretty intimidating and I can't find the relevant bits for me
</blockquote></details>

Done refactoring. See my reply to another comment of yours below. I believe the code structure is much cleaner now, in terms of the separation of ML / data / UI code.

---

*[lstm-text-generation/package.json, line 12 at r3](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LFrvspT3Ar1dHSy36sE:-LFyDMid1RYJo7T0avDl:bzd54pj) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/6530029a52a19b79a787df3e96762974fe5afb40/lstm-text-generation/package.json#L12)):*
<details><summary><i>Previously, nsthorat (Nikhil Thorat) wrote…</i></summary><blockquote>

use ~ here and below
</blockquote></details>

`^` is what we are using for the other examples in this repo. Let's use `^` here for consistency. We can do batch replacement later.

---

*[lstm-text-generation/nn-text-generator.js, line 30 at r3](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LFsnbg74IsMdYZZRqG8:-LFxVh8wAxc0nxOPXSes:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/6530029a52a19b79a787df3e96762974fe5afb40/lstm-text-generation/nn-text-generator.js#L30)):*
<details><summary><i>Previously, nsthorat (Nikhil Thorat) wrote…</i></summary><blockquote>

we have a util.shuffle: https://github.com/tensorflow/tfjs-core/blob/master/src/util.ts#L23
</blockquote></details>

Done.

---

*[lstm-text-generation/nn-text-generator.js, line 55 at r3](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LFsniR03rv_7ImCdHBj:-LFxh9Ey0VJ9_rODIf3k:b385zn6) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/6530029a52a19b79a787df3e96762974fe5afb40/lstm-text-generation/nn-text-generator.js#L55)):*
<details><summary><i>Previously, nsthorat (Nikhil Thorat) wrote…</i></summary><blockquote>

tf.multinomial?
</blockquote></details>

Thanks for bringing up that function. I tried it. It seems to give vastly different results from the custom `sampleOneFromMultinomial` function here, to the extent that if I use tf.multinomial, the generated text don't make resemble the training text at all. I think this may have to do with the way I'm using tf.multinomial: I call it with the second argument (number of samples) set to 1, repeatedly in a loop. Do you know whether that's related the difference in behavior that I'm observing? In any case, I added a TODO item here to investigate that.

---

*[lstm-text-generation/nn-text-generator.js, line 78 at r3](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LFsnpePA7IZgBneaM2L:-LFyAVAo9goO0heM_dqp:b60dxch) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/6530029a52a19b79a787df3e96762974fe5afb40/lstm-text-generation/nn-text-generator.js#L78)):*
<details><summary><i>Previously, nsthorat (Nikhil Thorat) wrote…</i></summary><blockquote>

So the way we've been trying to structure these examples is to put the meat of the ML inside of the index.js file. When people look to see how to make their own generative text model they'll eventually end up here after finding the call site in index.js. The idea would be for them to easily transplant the example, understanding how the pieces of it works, however this file seems like it's generally useful as a neural network text generator utility (capturing corner cases, etc). I wonder if it's better to simply have much "stupider" code that is much easier to parse. As it is, even I'm having a hard time following some of this.

Few comments inline where I think it could be simplified:
</blockquote></details>

Done refactoring. Now we have four main JavaScript files:

- index.js: contains the meat of the machine learning code: model creation, training, inference + save/load code. Note that the save/load code is separated into a subclass for clarity.
- data.js: contains the `TextData` class that takes care of things like converting text string to one-hot encoding and drawing random slices from the text data.
- ui.js: sets up the UI logics
- utils.js: contains utility functions for sampling from probability distributions output by the model.

---

*[lstm-text-generation/nn-text-generator.js, line 122 at r3](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LFsp6Bu5lRt9GOSMdsB:-LFyBBxp-sSUPVQ7t3Vq:b-qkhjg3) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/6530029a52a19b79a787df3e96762974fe5afb40/lstm-text-generation/nn-text-generator.js#L122)):*
<details><summary><i>Previously, nsthorat (Nikhil Thorat) wrote…</i></summary><blockquote>

if you un-class-ify this, you dont need all the code for these getters
</blockquote></details>

As discussed offline, I feel this data-handling class is complex enough to merit an OOP approach.

---

*[lstm-text-generation/nn-text-generator.js, line 154 at r3](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LFsqZu9EAHtchv2GTwy:-LFyBJK463JifvWs3YC3:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/6530029a52a19b79a787df3e96762974fe5afb40/lstm-text-generation/nn-text-generator.js#L154)):*
<details><summary><i>Previously, nsthorat (Nikhil Thorat) wrote…</i></summary><blockquote>

this could be in a data.js file
</blockquote></details>

Done.

---

*[lstm-text-generation/nn-text-generator.js, line 206 at r3](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LFsqfZC4lO6_INve1Z8:-LFyE8UUDIPR5OSj2f2n:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/6530029a52a19b79a787df3e96762974fe5afb40/lstm-text-generation/nn-text-generator.js#L206)):*
<details><summary><i>Previously, nsthorat (Nikhil Thorat) wrote…</i></summary><blockquote>

you can make this simpler if this goes in index.js and calls an explicit "ui.update()" or something like that -- the genericness of this callback will confuse people
</blockquote></details>

Done.

---

*[lstm-text-generation/nn-text-generator.js, line 220 at r3](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LFsqnBw5hq1cNVI59hg:-LFyBP9WB_U9k1UX6Obu:bfoo0v8) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/6530029a52a19b79a787df3e96762974fe5afb40/lstm-text-generation/nn-text-generator.js#L220)):*
<details><summary><i>Previously, nsthorat (Nikhil Thorat) wrote…</i></summary><blockquote>

This method should be one of the first thing people see inside of index.js when they come to this example
</blockquote></details>

Yep. Good point. Done as you said.

---

*[lstm-text-generation/nn-text-generator.js, line 230 at r3](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LFsqwRu07A755YQJbwl:-LFyBS26-BQdwWS8X6xD:bqls7o2) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/6530029a52a19b79a787df3e96762974fe5afb40/lstm-text-generation/nn-text-generator.js#L230)):*
<details><summary><i>Previously, nsthorat (Nikhil Thorat) wrote…</i></summary><blockquote>

I don't think this example should have any of the logic for loading and saving models, it just complicates it. Think of examples as just explaining a single concept. Once you remove all this logic for saving, this becomes much much simpler and easier to grok
</blockquote></details>

As I replied above, the model save/load code is separated out from the ML code.

I feel this example can benefit from the save/load code, for the following reasons

- The training of this LSTM-based model is time-consuming enough that it would be unwieldy to throw away all the training result each time the browser tab is closed. I feel it's much more satisfying and efficient to be able to save the result of the training and resume training/inference later.
- Of course, we could let the example load pre-trained models. But that kind of defeats the purpose of showing non-trivial training in the browser in this example.
- Right now we have only one example in tfjs-example for the tf.Model save/load API. I think it would be nice to add another one.

---

*[lstm-text-generation/nn-text-generator.js, line 259 at r3](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LFsr6SlEYW00IUhQfDS:-LFyCokWARIgkYiwJMls:bdyfwok) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/6530029a52a19b79a787df3e96762974fe5afb40/lstm-text-generation/nn-text-generator.js#L259)):*
<details><summary><i>Previously, nsthorat (Nikhil Thorat) wrote…</i></summary><blockquote>

this error throwing should be at the UI level, not complicating this method
</blockquote></details>

Good point. Done. Same for other methods in this class.

---

*[lstm-text-generation/nn-text-generator.js, line 270 at r3](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LFsrDQN1vLlH77BWzRD:-LFyUHCXCU42RlrjRMln:bu5hzdg) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/6530029a52a19b79a787df3e96762974fe5afb40/lstm-text-generation/nn-text-generator.js#L270)):*
<details><summary><i>Previously, nsthorat (Nikhil Thorat) wrote…</i></summary><blockquote>

when you move this, it will become something like ui.getCharSetSize(), just like we do here: https://github.com/tensorflow/tfjs-examples/blob/master/webcam-transfer-learning/index.js#L80
</blockquote></details>

This is now TextData.charSetSize().

---

*[lstm-text-generation/nn-text-generator.js, line 282 at r3](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LFsrPUT-7fbO7gBXrUF:-LFyCswZ1Pk0Lesbue46:b80uczy) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/6530029a52a19b79a787df3e96762974fe5afb40/lstm-text-generation/nn-text-generator.js#L282)):*
<details><summary><i>Previously, nsthorat (Nikhil Thorat) wrote…</i></summary><blockquote>

compile should live alongside creating the model
</blockquote></details>

The reason why I separate `compile` out is the user may click the "Train model" button multiple times in a browser session and train the model multiple times with different learning rates. When you change the learning rate (or any other aspects of the optimizer), the `compile` methods has to be called again.

---

*[lstm-text-generation/nn-text-generator.js, line 291 at r3](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LFsrRl_3Jl05tmPdgTc:-LFyD2oVB4Ych_JCZ5Yy:b-usn1c8) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/6530029a52a19b79a787df3e96762974fe5afb40/lstm-text-generation/nn-text-generator.js#L291)):*
<details><summary><i>Previously, nsthorat (Nikhil Thorat) wrote…</i></summary><blockquote>

these methods should go away (remove and save), same with checking status
</blockquote></details>

See my comment on the model save/load code above.

---

*[lstm-text-generation/nn-text-generator.js, line 362 at r3](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LFsrWJu8v98MUEKmFfv:-LFyD74O1e3q5CG0HQ26:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/6530029a52a19b79a787df3e96762974fe5afb40/lstm-text-generation/nn-text-generator.js#L362)):*
<details><summary><i>Previously, nsthorat (Nikhil Thorat) wrote…</i></summary><blockquote>

ui.batchEnd() or something along those lines
</blockquote></details>

Done.

---

*[lstm-text-generation/nn-text-generator.js, line 384 at r3](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LFsr_plDaZWZx6ihfZs:-LFyD8k21fG1V6-NLvQm:b-rlaau7) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/6530029a52a19b79a787df3e96762974fe5afb40/lstm-text-generation/nn-text-generator.js#L384)):*
<details><summary><i>Previously, nsthorat (Nikhil Thorat) wrote…</i></summary><blockquote>

Update the UI instead of the console?
</blockquote></details>

This method is actually not needed given onBatchEnd. Removed it.

---

*[lstm-text-generation/nn-text-generator.js, line 423 at r3](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LFsrhkiEn9fiWUL4ncB:-LFyDDupBYbnVLbqKBDo:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/6530029a52a19b79a787df3e96762974fe5afb40/lstm-text-generation/nn-text-generator.js#L423)):*
<details><summary><i>Previously, nsthorat (Nikhil Thorat) wrote…</i></summary><blockquote>

this could live in a data.js file, same with all the other data-related stuff
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-:-LFyTTFADXEp4lNhHucK:b-bgrthq)*

nsthorat:
:lgtm_strong:

---

Review status: :shipit: complete! 1 of 1 LGTMs obtained

---

*[lstm-text-generation/nn-text-generator.js, line 55 at r3](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LFsniR03rv_7ImCdHBj:-LG0jFErBd8eO6n2D_Ei:b-9ee7wa) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/6530029a52a19b79a787df3e96762974fe5afb40/lstm-text-generation/nn-text-generator.js#L55)):*

Thanks for bringing up that function. I tried it. It seems to give vastly different results from the custom `sampleOneFromMultinomial` function here, to the extent that if I use tf.multinomial, the generated text don't make resemble the training text at all. I think this may have to do with the way I'm using tf.multinomial: I call it with the second argument (number of samples) set to 1, repeatedly in a loop. Do you know whether that's related the difference in behavior that I'm observing? In any case, I added a TODO item here to investigate that.
</blockquote></details>

That is quite strange -- are you sure you are using the right "normalized" flag?

---

*[lstm-text-generation/nn-text-generator.js, line 282 at r3](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LFsrPUT-7fbO7gBXrUF:-LG0kurf0gvPkBl4oBPp:bjjo27) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/6530029a52a19b79a787df3e96762974fe5afb40/lstm-text-generation/nn-text-generator.js#L282)):*

The reason why I separate `compile` out is the user may click the "Train model" button multiple times in a browser session and train the model multiple times with different learning rates. When you change the learning rate (or any other aspects of the optimizer), the `compile` methods has to be called again.
</blockquote></details>

Got it. It's probably fine to rebuild a model and recompile it every time for simplicity, but this is fine.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-:-LG0lD4yEEpqWWN5HFxy:b-jbsf8z)*

nsthorat:
@dsmilkov @tafsiri want to take a pass?

dsmilkov:
Took a quick pass. Just couple high-level comments

---

Reviewed 1 of 7 files at r1.
Review status: :shipit: complete! 1 of 1 LGTMs obtained

---

*[lstm-text-generation/data.js, line 25 at r4](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LG0ldUt8aaEQZnAx_FI:-LG0ldUt8aaEQZnAx_FJ:byeq37t) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/b3bc3c2c7be93221418c4338d57b224740a607a4/lstm-text-generation/data.js#L25)):*
> ```JavaScript
>  * This class manages the following:
>  *
>  * - Converting training data (as a string) into one hot-encoded vectors.
> ```

one-hot encoded (instead of one hot-encoded)

---

*[lstm-text-generation/data.js, line 34 at r4](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LG0oZvP2cV56Igi8_u4:-LG0oZvP2cV56Igi8_u5:b-rvpqfl) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/b3bc3c2c7be93221418c4338d57b224740a607a4/lstm-text-generation/data.js#L34)):*
> ```JavaScript
>    *
>    * @param {string} dataIdentifier An identifier for this instance of TextData.
>    * @param {string} textString The training test data.
> ```

is it the train or the test data? ('training test data' is a bit confusing)

---

*[lstm-text-generation/data.js, line 78 at r4](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LG0o1Wf3Gwc8zPKVIYp:-LG0o1Wf3Gwc8zPKVIYq:b-oe1eej) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/b3bc3c2c7be93221418c4338d57b224740a607a4/lstm-text-generation/data.js#L78)):*
> ```JavaScript
>    * Get the length of each training example.
>    */
>   sampleLen() {
> ```

no reason to have private methods and getters/setters for an example. Makes more sense for a library/something that is reused across multiple examples

---

*[lstm-text-generation/README.md, line 6 at r4](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LG0ooB71Ip95-5fyfKr:-LG0ooB71Ip95-5fyfKs:b-e65af7) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/b3bc3c2c7be93221418c4338d57b224740a607a4/lstm-text-generation/README.md#L6)):*
> ```Markdown
> 
> This example illustrates how to use TensorFlow.js to train a LSTM model to
> generate random text based on the patterns in a training text corpus such as
> ```

(no need to say "training"): ... on the patterns in a text corpus

---

*[lstm-text-generation/README.md, line 28 at r4](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LG0pSaA5QD2iIQC_xaH:-LG0pSaB6czMM4h1KvbC:b9cztc8) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/b3bc3c2c7be93221418c4338d57b224740a607a4/lstm-text-generation/README.md#L28)):*
> ```Markdown
> This example also illustrates how to save a trained model in the browser's
> IndexedDB using TensorFlow.js's
> [model saving API](https://js.tensorflow.org/tutorials/model-save-load.html),
> ```

I think model saving/loading could be a separate example. No need to complicate an already involved one with serialization. WDYT?

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-:-LG0rDvkDemGqylUuGW8:b-60i14a)*

tafsiri:
This is cool. I look forward to playing with it myself. Left a couple of code style comments. I would say i'm overall pro keeping the saving and loading code given @caisq' comments. Where I could see simplifying the code is to drop the SaveableLSTMTextGenerator subclass and just have some simple save and load methods in the main LTSMTextGenerator class (or even just free floating utility functions). I think the subclass does start to make things seem heavy. 

---

Review status: :shipit: complete! 1 of 1 LGTMs obtained

---

*[lstm-text-generation/data.js, line 45 at r4](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LG0pmYp9wO37fwDV1cJ:-LG0pmYp9wO37fwDV1cK:bpw819n) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/b3bc3c2c7be93221418c4338d57b224740a607a4/lstm-text-generation/data.js#L45)):*
> ```JavaScript
>     }
> 
>     this._dataIdentifier = dataIdentifier;
> ```

General code style comment for here and else where, i don't think you need the underscore prefix. You could just do this.dataIdentifier = dataIdentifier

---

*[lstm-text-generation/data.js, line 62 at r4](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LG0pveH2qPrClZnD5iv:-LG0pveH2qPrClZnD5iw:b-wudqpy) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/b3bc3c2c7be93221418c4338d57b224740a607a4/lstm-text-generation/data.js#L62)):*
> ```JavaScript
>    * @returns {string} The data identifier.
>    */
>   dataIdentifier() {
> ```

I would suggest removing these simple getters (that don't do any special processing) and just accessing the variable directly. It would reduce the amount code and the vars are 'public' anyway.

---

*[lstm-text-generation/data.js, line 168 at r4](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LG0qcmr8Pb7UTUDskjU:-LG0qcmsEKKvfhofUiRV:b-the27y) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/b3bc3c2c7be93221418c4338d57b224740a607a4/lstm-text-generation/data.js#L168)):*
> ```JavaScript
>    * Get the set of unique characters from text.
>    */
>   _getCharSet() {
> ```

again i don't think the underscore convention for private members is common in JS. google js styleguide suggests trailing underscores for private members. 

---

*[lstm-text-generation/README.md, line 7 at r4](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LG0riYK4IhI22K-qPap:-LG0riYK4IhI22K-qPaq:b-bwhaa7) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/b3bc3c2c7be93221418c4338d57b224740a607a4/lstm-text-generation/README.md#L7)):*
> ```
> Nietzsche's writing and the source code
> ```

would change to Nietzsche's writing *or* the source code, the and made me think they had been combined :)

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-:-LG0s6QRAl5-UnGntXAE:b-ksh5tv)*

caisq:
Review status: :shipit: complete! 1 of 1 LGTMs obtained

---

*[lstm-text-generation/data.js, line 25 at r4](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LG0ldUt8aaEQZnAx_FI:-LG0vKei5Yv_jGHeaFqr:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/b3bc3c2c7be93221418c4338d57b224740a607a4/lstm-text-generation/data.js#L25)):*
<details><summary><i>Previously, dsmilkov (Daniel Smilkov) wrote…</i></summary><blockquote>

one-hot encoded (instead of one hot-encoded)
</blockquote></details>

Done.

---

*[lstm-text-generation/data.js, line 34 at r4](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LG0oZvP2cV56Igi8_u4:-LG0vOAhBkASeAu8qI_F:bs3ytdf) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/b3bc3c2c7be93221418c4338d57b224740a607a4/lstm-text-generation/data.js#L34)):*
<details><summary><i>Previously, dsmilkov (Daniel Smilkov) wrote…</i></summary><blockquote>

is it the train or the test data? ('training test data' is a bit confusing)
</blockquote></details>

Ah - that's a typo. I did s/test/text.

---

*[lstm-text-generation/data.js, line 45 at r4](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LG0pmYp9wO37fwDV1cJ:-LG0vS8J1dSSdC-2cEOA:bojlqd1) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/b3bc3c2c7be93221418c4338d57b224740a607a4/lstm-text-generation/data.js#L45)):*
<details><summary><i>Previously, tafsiri (Yannick Assogba) wrote…</i></summary><blockquote>

General code style comment for here and else where, i don't think you need the underscore prefix. You could just do this.dataIdentifier = dataIdentifier
</blockquote></details>

I use the underscore prefix to mark private members. I think it is good practice to make public member distinct from private ones.  But you are right (as below) that Google JS style recommends trailing underscores instead of leading ones. I changed all these private member names to trailing underscore.

---

*[lstm-text-generation/data.js, line 62 at r4](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LG0pveH2qPrClZnD5iv:-LG0vuFa71jrjBt2vPo9:bqpwiab) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/b3bc3c2c7be93221418c4338d57b224740a607a4/lstm-text-generation/data.js#L62)):*
<details><summary><i>Previously, tafsiri (Yannick Assogba) wrote…</i></summary><blockquote>

I would suggest removing these simple getters (that don't do any special processing) and just accessing the variable directly. It would reduce the amount code and the vars are 'public' anyway.
</blockquote></details>

I think the getters have benefits: they reduce the likelihood of clients accidentally overwriting those values.

---

*[lstm-text-generation/data.js, line 78 at r4](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LG0o1Wf3Gwc8zPKVIYp:-LG0yXEd4z-XjQq31VFN:b1i7ky8) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/b3bc3c2c7be93221418c4338d57b224740a607a4/lstm-text-generation/data.js#L78)):*
<details><summary><i>Previously, dsmilkov (Daniel Smilkov) wrote…</i></summary><blockquote>

no reason to have private methods and getters/setters for an example. Makes more sense for a library/something that is reused across multiple examples
</blockquote></details>

See my reply to Yannick's comment above. I think getters make it safer. Also this is in data.js instead of index.js. So I think it's all right to have some additional boilerplate here for the safety.

---

*[lstm-text-generation/data.js, line 168 at r4](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LG0qcmr8Pb7UTUDskjU:-LG0w5l-4s0C5XYdTSyA:b-t1ksbq) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/b3bc3c2c7be93221418c4338d57b224740a607a4/lstm-text-generation/data.js#L168)):*
<details><summary><i>Previously, tafsiri (Yannick Assogba) wrote…</i></summary><blockquote>

again i don't think the underscore convention for private members is common in JS. google js styleguide suggests trailing underscores for private members. 
</blockquote></details>

OK. I changed the leading underscores to trailing underscores.

---

*[lstm-text-generation/README.md, line 6 at r4](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LG0ooB71Ip95-5fyfKr:-LG0xk251g-Cc6JddLMu:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/b3bc3c2c7be93221418c4338d57b224740a607a4/lstm-text-generation/README.md#L6)):*
<details><summary><i>Previously, dsmilkov (Daniel Smilkov) wrote…</i></summary><blockquote>

(no need to say "training"): ... on the patterns in a text corpus
</blockquote></details>

Done.

---

*[lstm-text-generation/README.md, line 7 at r4](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LG0riYK4IhI22K-qPap:-LG0x5BYDe5z7OYuh0Q-:bmfxgaw) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/b3bc3c2c7be93221418c4338d57b224740a607a4/lstm-text-generation/README.md#L7)):*
<details><summary><i>Previously, tafsiri (Yannick Assogba) wrote…</i></summary><blockquote>

> ```
> Nietzsche's writing and the source code
> ```

would change to Nietzsche's writing *or* the source code, the and made me think they had been combined :)
</blockquote></details>

Done :)

---

*[lstm-text-generation/README.md, line 28 at r4](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LG0pSaA5QD2iIQC_xaH:-LG0xkq66pR6X8uGEUKP:b5zgqrp) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/b3bc3c2c7be93221418c4338d57b224740a607a4/lstm-text-generation/README.md#L28)):*
<details><summary><i>Previously, dsmilkov (Daniel Smilkov) wrote…</i></summary><blockquote>

I think model saving/loading could be a separate example. No need to complicate an already involved one with serialization. WDYT?
</blockquote></details>

See my reply to Nikhil's previous comment. I think there are good reasons to show saving/loading in this example. If we are worried about the code being too big for a blog post or a book chapter, the concern is alleviated by my splitting of the save/load code into a subclass.

---

*[lstm-text-generation/nn-text-generator.js, line 55 at r3](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LFsniR03rv_7ImCdHBj:-LG0yDXA1qcWKo_jXGIt:b-oq1c5u) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/6530029a52a19b79a787df3e96762974fe5afb40/lstm-text-generation/nn-text-generator.js#L55)):*
<details><summary><i>Previously, nsthorat (Nikhil Thorat) wrote…</i></summary><blockquote>

That is quite strange -- are you sure you are using the right "normalized" flag?
</blockquote></details>

Ah - good point. I didn't set it to `true`. Everything is working after I set it to `true`. Done.

---

*[lstm-text-generation/nn-text-generator.js, line 282 at r3](https://reviewable.io/reviews/tensorflow/tfjs-examples/103#-LFsrPUT-7fbO7gBXrUF:-LG0yK1d9QaLgjYyXh2_:b-o6maf1) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/6530029a52a19b79a787df3e96762974fe5afb40/lstm-text-generation/nn-text-generator.js#L282)):*
<details><summary><i>Previously, nsthorat (Nikhil Thorat) wrote…</i></summary><blockquote>

Got it. It's probably fine to rebuild a model and recompile it every time for simplicity, but this is fine.
</blockquote></details>

Calling compile() repeatedly is pretty standard in Python Keras code as well.

---


*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/103)*

brannondorsey:
Nice work. I recently implemented that same Keras example in tfjs, although I didn't take the time to create a nice UI out of it. I've had a fair amount of experience with char-rnn style text generation LSTMs in various frameworks, etc. and one thing I continue to find odd is that most of the examples out there do not use validation data. Its pretty easy to fit these types of models and you usually get nice loss curves, but without also tracking validation loss you have no idea when to stop training. With this text based stuff there is often a fine line between learning and memorizing, and without a peek at validation loss you have no clue where that line actually is. Would you be open to including a validation split in this example and graphing val_loss as well?

caisq:
@brannondorsey Your suggestion about adding validation makes a lot of sense. I was thinking of it at the back of my head, too. Done. The validation split defaults to 0.0625 in the UI. During training, the training and validation loss values are plotted in the same axes.

brannondorsey:
:clap::clap::clap: amazing! Thanks for adding this ;)

Any reasoning behind 0.0625 as a default split? That sounds a bit on the low end to me but ¯\_(ツ)_/¯. I usually use a val split between 0.1 and 0.2 depending on how much training data I have.

I was playing with your example today and its pretty nice. I did find myself curious about wanting to upload my own text files to use as training data. Have you thought about adding that feature in addition to Nietzsche and the tfjs source code? I could see people using this as a general purpose char-rnn demo tool if that functionality were provided.

caisq:
@brannondorsey Again, good question.  Yes, validationSplit is usually set to a value of about 0.1. But in this case, I didn't find it to make any substantial difference whether it's >0.1 or 0.0625. The magical-looking number 0.0625 comes from the fact that by default, there are 16 batches in the data for an epoch. 0.0625 is just one batch. We don't have tons of data in the browser (the two datasets are about 500kB each). I don't want to "waste" too much data on validation, especially because in this case, the default model size doesn't lead to much overfitting (i.e., the default model size is not powerful enough to accurately memorize the training data).

As for arbitrary input data, I think it's a great idea. However, since the review process is already getting a little long and the code change getting a little too much to track, I'd prefer checking in this PR first and then add the support for user-supplied dataset in a followup PR. 

Are you interested in helping out @brannondorsey ? If not, I'll be happy to do it myself. 

Let me know if you have any remaining thoughts,  @nsthorat @dsmilkov @tafsiri

brannondorsey:
Congrats on the merge :clap:. Sorry for the delayed reply, I've been moving to the east coast the last few days and things have been a bit nuts. @caisq I'd be happy to add this "file upload" functionality if you haven't already started on it.

caisq:
@brannondorsey Good to hear! I haven't started on that yet. Looking forward to reviewing your PR! 

One thought: the user-defined text can probably be saved in browser local storage.

brannondorsey:
Sounds good. Not sure of an ETA yet as my dev machine is still in boxes, but I should be able to put some time towards it in the next week or so. 

Local storage is definitely the right choice. Don't worry, I plan to keep this functionality 100% client side ;)

PR #105, opened 2018-07-03
Upgrade tfjs from 0.11 to 0.12 in examples

caisq:
Also change `^` to `~` in the version strings, so that upgrades like this won't be absolutely necessary in the future.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/105)

### DIFF
--- a/addition-rnn/package.json
+++ b/addition-rnn/package.json
@@ -9,7 +9,7 @@
     "node": ">=8.9.0"
   },
   "dependencies": {
-    "@tensorflow/tfjs": "~0.12.0",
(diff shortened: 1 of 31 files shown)


dsmilkov:
I made them ^ so that we are "forced" to keep our examples up to date with our API. Using `~` will take the examples much more stable, but then we can be lazy about upgrading them

---

Review status: :shipit: complete! 1 of 1 LGTMs obtained

---



*Comments from [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/105#-:-LGWcDnJ78MbxDoFi6S2:bmjyder)*

PR #108, opened 2018-07-09
[WIP] Multi Linear Regression Example

ManrajGrover:
This PR adds multivariate linear regression example.


ManrajGrover:
Hi @caisq,

This looks ready for first review.

ManrajGrover:
Hi Stanley,

> it looks like this is designed for node. Is it possible to adjust this to instead focus on the browser?

Would this example be completely browser focused?

> A pattern for loading data without a file system can be cloned from the mnist example

I'll have a look but current dataset contains `csv` and not images

> Technically, the normalization parameters are learned parameters. They should be estimated from the training set and applied to the test set.

Agreed, will fix this

> would Date().getTime(); work for these?

For web, we have [`performance.now()`](https://developer.mozilla.org/en-US/docs/Web/API/Performance/now) which would be a better option

ManrajGrover:
Made the changes

Apologies, I'm not comfortable with permissions asked by Reviewable and hence don't make use of it.

> would Date().getTime(); work for these?

Done

Made use of `performance.now()`

> any reason not to use the layers API? will make the model definition much simpler

There is already an example for regression making use of Layers API.

> I think it's fine to keep as is -- none of the other examples do this.

Okay

> +1 -- they don't necessarily have to be learned but the normalization parameters should span both training and test set (computing mean / variance across the entire dataset).

I feel we should compute mean and variance on the training set and apply it to testing to avoid any leakage of information. I've made the changes accordingly. Please correct me if I'm wrong.

> use export const BostonHousingDataset

Done

> no need for undescores

Done

> put this in utils

Done

> put this in and readCsv in utils (since their API is pretty easy to understand)

Done

> license at the top of thise file

Done

> you cant just export const stddev? why module.exports?

I'm not aware of nodejs version for which examples should be compatible and hence went with it

ManrajGrover:
@caisq Made the changes

ManrajGrover:
> --> Validation. We will use "validation" and "test" to mean different things. "Validation" is the held-out data that we run evaluation on during training. It is used for things like deciding when to stop training.

@caisq In this case, although model is being used to predict on test data after every epoch, early stopping is not being decided on it. It is being decided based on validation data which I assume `model.fit` is handling internally (since it accepts validation split as a param) Hence, model's learning is not being influenced by test data. Does `model.fit` return validation loss? I plan to plot it along with training loss (which is one of the reason I had moved test prediction inside callback)

> "Test" is the held-out data that the model never sees during training. It is used only after the training completes.

I'll move it out of the callback.

caisq:
@ManrajGrover Not sure I fully understood your last comment. But to answer your question: yes, model.fit() returns loss values for both the training and validation sets if validatoinSplit is set to a value >0. For example, `const history = await model.fit(...)` will return a `history` object that contains 'loss' and 'val_loss' in its 'history' field, which can be plotted. The other way to get the validation loss is to use the callbacks and use `logs.val_loss`, as you are currently doing.

I actually think getting the validation loss from the return value of model.fit() is better than getting it from the callbacks, because of simpler code. This is the 2nd concrete code example of the teaching material and we are probably not quite ready to dive into the details of callbacks yet.

In any case, I think plotting val_loss alongside loss is important because it shows the important concept of underfitting and overfitting. Running `model.evaluate()` on a separate test set to get the test loss value will also be nice, because that's the most objective evaluation of the model's accuracy.

caisq:
@ManrajGrover Yep. Sorry what I wrote before might be misleading. Early stopping is not done in your code. To do early stopping, you need a callback like https://keras.io/callbacks/#earlystopping, which is not implemented in tensorflow.js yet. For this example, showing plots of training and validation losses (along with a test loss value) suffices. There is no need to actually do early stopping.

caisq:
Looking forward to your next revision before I can approve this PR. Thanks.

ManrajGrover:
@caisq Done

PR #110, opened 2018-07-10
MNIST Node: Fix bugs in generateBatch

manrajgrover:
This PR fixes batched data problem.

1. For both test and train, train data was being used
2. Fixes edge case when batch size> last batch

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/110)


nsthorat:
cc @nkreeger

PR #113, opened 2018-07-14
small spelling mistake

tristansokol:
logits -> digit

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/113)

### DIFF
--- a/mnist-core/model.js
+++ b/mnist-core/model.js
@@ -101,7 +101,7 @@ export function predict(x) {
   return Array.from(pred.dataSync());
 }
 
-// Given a logits or label vector, return the class indices.
+// Given a digit or label vector, return the class indices.
 export function classesFromLabel(y) {
   const axis = 1;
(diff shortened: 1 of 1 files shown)


nsthorat:
`y` actually in this case should be the logits or the label vector. Just curious -- why do you think this is a mistake?

tristansokol:
Oh sorry. I had never heard of the word `logits` and didn't encounter it anywhere else in the source code or in the html example: https://js.tensorflow.org/tutorials/mnist.html

Do you have a good definition for what `logits` means? The [wiki](https://en.wikipedia.org/wiki/Logit) definition doesn't really seem transposable with `label vector`

nsthorat:
Check out the definition here: https://developers.google.com/machine-learning/glossary/#logits

Logits typically are the unnormalized predictions, right before the softmax normalization function.

PR #115, opened 2018-07-17
Support user-supplied text data; configurable text data urls;

loretoparisi:
Added
- Support for user-supplied text data;
- configurable text data urls in the html select

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/115)


caisq:
FYI, @brannondorsey

brannondorsey:
:clap: bravo @loretoparisi, thanks for the addition. @caisq, been a bit underwater with work & moving recently, sorry my changes never made it to master. Glad @loretoparisi took the reigns.

loretoparisi:
@brannondorsey you are welcome! 

I was thinking to a further improvement using web workers to distribute the load over processes, but maybe this could be another example, if you think it could be helpful. I did something like that here: in PAIR code for dataset rendering via D3 that could be a CPU intensive task, https://github.com/PAIR-code/facets/commit/7378f027f919bd0caa1ec3c7a22633f2a36b0741

I was thinking to this from the beginning, but I was aware of some issues in the use of webGL, as stated here https://github.com/tensorflow/tfjs/issues/102

PR #117, opened 2018-07-23
multivariate linear regression (Boston Housing): Some cleanups and refactoring

caisq:
- Use official GCS URLs for the datasets
- Simplify the plotting code in index.js by pushing some logic to ui.js
- Simplify `ui.updateStatus`
- Improve the axis labeling of the loss plot
- Improve the final loss summary message: include all three: train-set, validation-set and test-set losses.

This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/117)


caisq:
cc @ManrajGrover

multivariate-linear-regression/utils.js
```diff
@@ -49,7 +45,7 @@ export const loadCsv = async (filename) => {
   return new Promise(resolve => {
     const url = `${BASE_URL}${filename}.csv`;
 
-    console.log(`  * Downloading from: ${url}`);
+    console.log(`  * Downloading data from: ${url}`);
```
ManrajGrover: This was a debug statement. It can be removed now or we can update `status` accordingly.

multivariate-linear-regression/index.js
```diff
@@ -68,12 +61,12 @@ data.loadData().then(async () => {
...
-
+  await ui.updateStatus('Running on test data...');
   const result =
       model.evaluate(testData.data, testData.target, {batchSize: BATCH_SIZE});
-
   const testLoss = result.get().toFixed(4);
-
-  await ui.updateStatus('status', `Test set loss: ${testLoss}`);
+  await ui.updateStatus(
+      `Final train-set loss: ${trainLoss};` +
```
ManrajGrover: Probably we can break it into 3 different lines for improving readability. Also, we should round off `trainLoss` and `valLoss`.

caisq:
@ManrajGrover This is merged, which should unblock in in the NN regression work. Thanks!

PR #119, opened 2018-07-29
Classification Example: Adds Phishing Website URL Classifier

manrajgrover:
This PR adds Phishing Website URL Classifier example.

(Need to tune params)

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/119)


manrajgrover:
Hi @caisq,

> See my comment on index.js. For binary classification problems, the binaryCrossentropy loss function should be used. Therefore there is no need to one-hot encode the targets.

Agreed. Actually, there was another dataset on which I was testing this which had 3 labels and forgot to make this change. Thanks for catching it!

> This is a good opportunity to show how to calculate the following in tfjs:
>  Precision and recall
>  Maybe also ROC.

Yup, I'll add precision and recall soon. I'll look into ROC calculation since it is not as straightforward.

> Before we merge this PR, can you do some hyperparameter tuning. The test-set accuracy is currently about 95%. Can we get better than that?

On the current dataset split, I just performed tuning and got test accuracy close to 95.5%. Currently, I'm not happy with the data split since it is 50:50 right now and 0.2 of the training dataset is used for validation. I feel we should make it either 70:15:15 or 60:20:20. Please share your inputs on it.

> High-level comment: I looked at the training curves and it seems that neither the training or validation loss converged given the number of epochs. Can we make sure that the training converges?

I just hyper tuned the params and increased epochs as well but going for training convergence would make the model overfit to the training data.

> After the training finishes, is it possible to let the user enter an actual URL and get a prediction from the model? That would be very cool. It would also solve the slight shortcoming of this example: it doesn't show how to do inference with the trained model.
> 
> But unfortunately this looks like something too complicated for this simple demo, because some features, e.g., ON_MOUSEOVER, cannot be easily extracted.
>
> It may be good to think about whether we can make this into a Chrome extension - that would probably make a very good tfjs + browser extension demo (and potentially very useful in practice too).

Extraction of certain features used in the dataset would make it a little complex. Perhaps, we can have a look at feature importance and reduce the feature set based on feasibility without compromising on accuracy.

Working on rest of the changes and will push them soon.

manrajgrover:
> I think it makes sense to add the precision and recall metrics to TensorFlow.js and use them here, instead of rolling one-off custom code. That will help many others.
>
> Manraj, do you want to take up that task? You can model the new metrics functions after:
>
> https://www.tensorflow.org/api_docs/python/tf/metrics/precision

Would this be part of Core API or Layers API? Currently, I see metrics are implemented in Layers whereas TensorFlow Core also has some metrics implemented. We can merge this and fix it once `precision` and `recall` are part of tfjs.

Made rest of the changes.

caisq:
@manrajgrover Please implement it in tfjs-layers, alongside the other metrics like `binaryAccuracy` in `metrics.ts`. Thanks!!

ericdnielsen:
@manrajgrover Hi, I was just looking at this since I was asking Shanqing about recall/precision metrics and he pointed me here.  Wondering what your status was?

manrajgrover:
@ericdnielsen My apologies. I've been busy with team movements and new research work. I did see the TensorFlow implementation and found precision depends on a lot of [internal operations](https://github.com/tensorflow/tensorflow/blob/r1.10/tensorflow/python/ops/metrics_impl.py#L88) which are not currently implemented (which should be for these operations). I would need some time to get on this. If its urgent, probably someone else can take these up.

ericdnielsen:
@manrajgrover , Shanqing and I chatted, and we'll be ok with splitting the recall/precision work for the underlying API to a different PR, and getting this PR in with the current non-shared API version.

I'm not sure if there's any other outstanding issues on this PR.

manrajgrover:
@ericdnielsen This PR is ready to go. We can come back and replace current implementation of precision and recall with tfjs implementation.

I'll try to complete precision and recall over the weekend if its okay with you.

manrajgrover:
@caisq This examples now makes use of recently added `tf.metrics.precision` and `tf.metrics.recall`

PR #125, opened 2018-08-13
Improve mnist-node example

caisq:
Main changes:

- Simplify data.js: No need to shuffle data as Model.fit() shuffles training data by default
- Rename index.js as main.js
- Replace iterative, small-data-size Mode.fit() calls with a single Model.fit() call on all training data
- Increase the capacity of the model and add Dropout. The test accuracy is increased from about 0.987 to 0.996.
- Add progress bars during training.
- Update dependency versions

This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/125)


mnist-node/main.js
```diff
@@ -0,0 +1,82 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * =============================================================================
+ */
+
+const tf = require('@tensorflow/tfjs');
+const ProgressBar = require('progress');
+require('@tensorflow/tfjs-node');
```
nkreeger: nit: keep tfjs* packages together.

mnist-node/main.js
```diff
@@ -0,0 +1,82 @@
...
+  let epochBeginTime;
+  let millisPerStep;
+  const epochs = 20;
+  const batchSize = 128;
+  const validationSplit = 0.15;
+  const numTrainExamplesPerEpoch =
+      trainImages.shape[0] * (1 - validationSplit);
+  const numTrainBatchesPerEpoch =
+      Math.ceil(numTrainExamplesPerEpoch / batchSize);
+  await model.fit(trainImages, trainLabels, {
```
nkreeger: Minor question: when I wrote this - the internals to model's built-in shuffle was slow from all slicing. Is this still the case?

PR #130, opened 2018-08-14
Boston housing example updates to match book

bileschi:
Updates to code organization to facilitate the narrative in the book.

Major changes are:
* inclusion of baseline loss computation and UI element to hold it
* normalization code in TF instead of raw JS (renamed utils -> normalization)
* data.js no longer depends on tf

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/130)


boston-housing/index.js
```diff
@@ -47,44 +65,41 @@ export const linearRegressionModel = () => {
...
   model.add(tf.layers.dense({units: 50, activation: 'sigmoid'}));
   model.add(tf.layers.dense({units: 1}));
 
   return model;
 };
 
 /**
- * Fetches training and testing data, compiles `model`, trains the model
- * using train data and runs model against test data.
+ * Fetches training and testing data, normalizes, compiles `model`, trains the
```
manrajgrover: `run` no longer fetches or normalizes data. We can move this part `arraysToTensors`

boston-housing/index.js
```diff
@@ -94,17 +109,32 @@ export const run = async (model) => {
...
-  await data.loadData();
-  await ui.updateStatus('Data loaded!');
+  await bostonData.loadData();
+  ui.updateStatus('Data loaded, converting to tensors');
+  arraysToTensors();
+  ui.updateStatus(
+      'Data is now available as tensors.\n' +
+      'Click a train button to begin.');
+  ui.updateBaselineStatus('Estimating baseline loss');
+  await computeBaseline();
```
manrajgrover: `computeBaseline` does not return a promise. We shouldn't `await` it

boston-housing/data.js
```diff
@@ -15,78 +15,108 @@
...
   }
 
   /** Loads training and test data. */
   async loadData() {
-    this.dataset = await Promise.all([
-      utils.loadCsv(TRAIN_DATA), utils.loadCsv(TRAIN_TARGET),
-      utils.loadCsv(TEST_DATA), utils.loadCsv(TEST_TARGET)
-    ]);
-
-    let {dataset: trainDataset, vectorMeans, vectorStddevs} =
```
manrajgrover: I was not liking this part of code myself. Thanks for cleaning it up! :)

boston-housing/data.js
```diff
@@ -15,78 +15,108 @@
...
     this.testSize = 0;
-    this.trainBatchIndex = 0;
-    this.testBatchIndex = 0;
-
-    this.NUM_FEATURES = 12;
   }
 
   get numFeatures() {
-    return this.NUM_FEATURES;
+    return this.trainFeatures[0].length;
```
manrajgrover: Now that we are using `this.trainFeatures[0]` length as the number of features, I feel this will throw an error if `numFeatures` is accessed before calling `loadData`. We should either call `loadData` in the constructor or check if it is null before accessing the element.

boston-housing/index.js
```diff
@@ -47,44 +65,41 @@ export const linearRegressionModel = () => {
...
 
   let trainLoss;
   let valLoss;
   await ui.updateStatus('Starting training process...');
-  await model.fit(trainData.data, trainData.target, {
+  await model.fit(tensors.trainFeatures, tensors.trainTarget, {
     batchSize: BATCH_SIZE,
     epochs: NUM_EPOCHS,
-    validationSplit: 0.2,
+    validationData: [tensors.testFeatures, tensors.testTarget],
```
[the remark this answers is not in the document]
bileschi: I agree.  I've switched it back.  It complicates the book some as it introduces yet another new concept (i.e., the validation / testing distinction) but I think we can manage, and it's best to use best practices here where the code is very likely to be copied.

boston-housing/index.js
bileschi: Thanks for catching this!  I should go back through the doc to see if the comments are still correct.

boston-housing/index.js
bileschi: Done.

boston-housing/data.js
bileschi: :)

bileschi:
Thanks for the feedback and for putting together such a great example Manraj!  Everyone coming to learn and use TensorFlow.js is going to benefit from having such clear and instructive examples! 

My goal in this change is to adjust the code to clarify the narrative in chapter 2 of the book.  In some places I think it makes sense to do things in ways we wouldn't normally, just to facilitate the instruction and not distract the reader from topic at hand.

Thanks again!
Stan

boston-housing/data.js
bileschi: Agreed, updated the object to keep track of its load state in a private boolean member variable and to throw an error if the features are accessed before the data is loaded.

I've also removed the other two pieces of metadata, that weren't actually being used anymore (since tensor2d is able to determine the shape automatically from input type number[][]).

PR #134, opened 2018-08-30
Revert to @tensorflow/tfjs-node instead of the GPU package.

nkreeger:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/134)


caisq:
Another indication that we need some testing for the tfjs-examples repo.

PR #141, opened 2018-09-30
[boston-housing] Add a one-hidden-layer NN option; use leCunNormal initializer

caisq:
* Adding the one-hidden-layer NN fits the teaching material
* leCunNormal initailizer leads to more stable and on-average better
  training results from the one-hidden-layer and two-hidden-layer NNs.

  For example, see final test-set MSEs from:

  - 2 hidden layers, default kernel initializer:
    17.2148, 12.8895, 13.1877, 13.5261
  - 2 hidden layers, leCunNormal kernel initializer:
    12.1388, 11.4624, 13.4111, 12.1905

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/141)

### DIFF
--- a/boston-housing/index.js
+++ b/boston-housing/index.js
@@ -70,7 +70,7 @@ export function multiLayerPerceptronRegressionModel1Hidden() {
   model.add(tf.layers.dense({
     inputShape: [bostonData.numFeatures],
     units: 50,
-    // activation: 'sigmoid',
+    activation: 'sigmoid',
     kernelInitializer: 'leCunNormal'
   }));
   model.add(tf.layers.dense({units: 1}));
--- a/boston-housing/index.js
+++ b/boston-housing/index.js
@@ -51,9 +51,31 @@ export const arraysToTensors = () => {
  *
  * @returns {tf.Sequential} The linear regression model.
  */
-export const linearRegressionModel = () => {
+export function linearRegressionModel() {
   const model = tf.sequential();
   model.add(tf.layers.dense({inputShape: [bostonData.numFeatures], units: 1}));
+
+  model.summary();
+  return model;
+};
+
+/**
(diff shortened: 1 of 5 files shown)


caisq:
cc @manrajgrover

boston-housing/index.js
```diff
@@ -51,28 +51,56 @@ export const arraysToTensors = () => {
  *
  * @returns {tf.Sequential} The linear regression model.
  */
-export const linearRegressionModel = () => {
+export function linearRegressionModel() {
   const model = tf.sequential();
   model.add(tf.layers.dense({inputShape: [bostonData.numFeatures], units: 1}));
+
+  model.summary();
```
manrajgrover: Do we plan to show model summary in console? Can we show it in UI?

boston-housing/index.js
```diff
@@ -51,28 +51,56 @@ export const arraysToTensors = () => {
...
  *
  * @returns {tf.Sequential} The multi layer perceptron regression model.
  */
-export const multiLayerPerceptronRegressionModel = () => {
+export function multiLayerPerceptronRegressionModel1Hidden() {
+  const model = tf.sequential();
+  model.add(tf.layers.dense({
+    inputShape: [bostonData.numFeatures],
+    units: 50,
+    // activation: 'sigmoid',
```
manrajgrover: I was seeing similar results for linear regressor and MLP with 1 hidden layer. This should be uncommented to fix that.

boston-housing/index.js
caisq: For now it's okay. We'll switch to tf.vis when its model visualization component is ready.

boston-housing/index.js
caisq: Good catch. Done.

PR #147, opened 2018-10-09
Show best practice of shuffling data

caisq:


caisq:
FYI, this will likely be eventually replaced by shuffle() in the dataset API. But until that happens, let's do this in order to show best practice.

PR #160, opened 2018-10-23
Upgrade parcel

tafsiri:
This upgrades parcel in all of the examples so that sourcemaps will work correctly and allow easier debugging.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/160)


tafsiri:
cc @kangyizhang you may want to do something similar in the data demo.

PR #161, opened 2018-10-23
Upgrade parcel and babel

tafsiri:
making a new PR for this since the last one wasn't updating.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/161)


tafsiri:
@dsmilkov i reverted the change since it was unnecessary (getting started does not use parcel). @caisq will do once i can merge, looks like github may be having some issues atm.

PR #164, opened 2018-10-25
Integrate tfjs vis. refresh look and feel of examples: addition-rnn, cart-pole and mnist

tafsiri:
Here is a preview of the direction i'm thinking for unifying the look and feel of some of the examples and making them a bit more approachable. I'm doing this work as I integrate tfjs-vis.

Before doing code review I'd like to get an initial sense of whether we are happy with this direction, then I can make another comment to point out specific things to look at (including some changed behaviour). This preview has 3 of the examples done. I've included screenshots below. 

To give you some sense of what is underlying this, one of the main changes is to develop a stylesheet with a few basic layout helpers and styles that any examples can use (not all have to if they have a UI we are happy with). 

screen shot 2018-10-25 at 4 06 50 pm

screen shot 2018-10-25 at 4 08 00 pm

screen shot 2018-10-25 at 3 55 22 pm

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/164)


nsthorat:
Looks awesome! Let me know when you want me to look at code.

dsmilkov:
+1 This is great! Let us know when ready

tafsiri:
@caisq i see the dropdown in the screenshot above, it currently says "ConvNet", is there another one?

caisq:
@tafsiri Ah - it's kind of smallish that I missed it. Now I see it.

manrajgrover:
This looks amazing! :smile: Can review the code if needed.

tafsiri:
This is updated and ready for review. Here is a list of some of the editorial changes I made. These are changes that more reflect my opinion on how to simplify things and make the example easier to follow along. They may be up for debate so i want to point them out here.

- The 'number of predictions per second' chart has been removed from addition-rnn to simplify things (I think i generally have removed this metric where I saw it). 
- Cart pole example splits the various controls/display into 4 sections. I also set 'render during training' to default to true as i think it makes the example more interesting to look at. 
- Mnist: Run and display an interest test every ~10 batches, this shows the performance of the classifier improving in a more visual way. I also increased the batch size so that it would converge faster.

General implementation note. 

This PR introduces a `shared` folder what currently only contains a the shared css file and the other files used to generate it. Parcel based projects will directly include the css file and parcel will handle pulling it into the project bundle. For non parcel based projects there will be a script to copy the css file into those folders on build. I'll probably do those last to reduce PR diff size.

The plan is to update the shared template to make something that works for 80%+ of examples so the styling of these may change stlightly, but i'm confident that the current look and feel can work.

@manrajgrover Thanks! 😄 I'm excited to make these more inviting. Feel free to review code, though I'd say more importantly for me is feedback on the text and descriptions/instructions as the main goal in helping newcomers grok these examples more easily.

tafsiri:
@caisq Thanks for the review. I've addressed your comments.

PR #173, opened 2018-11-09
[simple-object-detection] Add files for inference

caisq:
- index.html and index.js
- style.css
- Add script for uploading trained model

### DIFF
--- a/simple-object-detection/README.md
+++ b/simple-object-detection/README.md
@@ -39,7 +39,10 @@ For example, the hosted model is trained with the 10000 examples, using
 the command line:
 
 ```sh
-yarn train --numExamples 10000
+yarn train \
+    --numExamples 20000 \
+    --initialTransferEpochs 100 \
+    --fineTuningEpochs 200
 ```
 
 See `train.js` for other adjustable parameters.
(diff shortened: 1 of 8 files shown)


bileschi:
:lgtm_strong:

tafsiri:
___
*[simple-object-detection/index.js, line 113 at r1](https://reviewable.io/reviews/tensorflow/tfjs-examples/173#-LRCyguqEl3KC1Jtdvyo:-LRH_IUCC5UvFVzeSSvN:b-aw2szk) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/62829571a728361dd00a8f553ad2941d6d84181e/simple-object-detection/index.js#L113)):*

Daniel and other have suggested that we prefer `data()` to `dataSync()`, because the former doesn't block the main thread while the latter does. It's not a big deal here because there isn't much UI interaction, but it's still good to show best practice.
</blockquote></details>

@caisq wrt the data/vs dataSync comment. My main point was that it looks this data was already fetched from GPU in line 112 (targets.dataSync()). I wasn't suggesting switching to using dataSync, but rather using the already downloaded values. Its not a big deal as it is cached anyway.

PR #176, opened 2018-11-14
[simple-object-detection] Allow some noise objects to obscure target object

caisq:
- Previously, all the background objects (line segments and circles) were drawn
  before the target objects and hence are in the background.
- In this PR, we allow 50% of the background objects to be drawn after the target
  objects. As a result, some of them will end up on to of the target objects and
  partially cover them.
- The model turns out to be able to handle it. There is an expected slight increase
  in the validation loss due to the partial covering, but the increase doesn't  lead
  to any serious degradation of the object classification or localization quality.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/176)

### DIFF
--- a/simple-object-detection/index.js
+++ b/simple-object-detection/index.js
@@ -104,7 +104,8 @@ async function runAndVisualizeInference(model) {
   const {images, targets} = await synth.generateExampleBatch(
       numExamples, numCircles, numLineSegments);
 
-  const boundingBoxArray = Array.from(targets.dataSync()).slice(1);
+  const targetsArray = await targets.data();
+  const boundingBoxArray = Array.from(targetsArray).slice(1);
   const t0 = tf.util.now();
   // Runs inference with the model.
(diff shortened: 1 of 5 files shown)


bileschi:
___
*[simple-object-detection/synthetic_images.js, line 158 at r1](https://reviewable.io/reviews/tensorflow/tfjs-examples/176#-LRHC6qQCGsTRBMGCJbB:-LRHC6qQCGsTRBMGCJbC:b71j32m) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/c297113ec2123104f66e24a302aeff52f26334ad/simple-object-detection/synthetic_images.js#L158)):*
> ```
> (let i = 0; i < numCircles / 2; ++i) 
> ```
You could rid yourself of the off-by-one counting issue for odd numbers of circles by replacing this with:
```
(let i = numCircles/2; i < numCircles; ++i)
```

PR #188, opened 2018-12-03
[For discussion] Upgrade @tensorflow/tfjs dependency: 0.13.3 --> 0.14.0; Fix dtype-related breakages

caisq:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/188)

### DIFF
--- a/addition-rnn/package.json
+++ b/addition-rnn/package.json
@@ -9,7 +9,7 @@
     "node": ">=8.9.0"
   },
   "dependencies": {
-    "@tensorflow/tfjs": "^0.13.3",
+    "@tensorflow/tfjs": "^0.14.0",
     "@tensorflow/tfjs-vis": "^0.2.0"
   },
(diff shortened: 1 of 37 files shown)


dsmilkov:
+1. We should also add it to release notes. This improved dtype inference is the right way forward, unfortunately it's breaking, but reverting it is not ideal.

dsmilkov:
Perhaps better is to release a patch version that relaxes the dtype constraint for binary_ops (add/sub/mul...) and matmul, like numpy does, which will unbreak users, and we wanted to do this anyhow.

dsmilkov:
I'm happy to do the relaxing of dtype in ops, EOD today or tomorrow, and shouldn't be much work.

caisq:
Closing this PR in favor of https://github.com/tensorflow/tfjs-core/pull/1432

PR #193, opened 2018-12-11
Update boston-housing example with tf.data

kangyizhang:
Update boston-housing example with tf.data according to https://github.com/tensorflow/tfjs-data/tree/master/demo/boston-housing

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/193)


bileschi:
Thanks!  Just a few suggestions.

caisq:
@kangyizhang Thanks for the PR. First off, can you let me know how many times is fitDataset() slower compared to the fit() in this boston-housing example?

PR #196, opened 2018-12-19
adds per-epoch timing to output UI

bileschi:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/196)

PR #203, opened 2019-01-04
Forks iris example into with and without tfjs-data versions

bileschi:
The with tfjs-data example is called `iris-fitDataset` and is identical to where master was after #197.  
The without tfjs-data example is called `iris` and is identical to where master was at  #196 .

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/203)


iris-fitDataset/data.js
```diff
@@ -0,0 +1,128 @@
...
+  [6.9, 3.1, 5.1, 2.3, 2], [5.8, 2.7, 5.1, 1.9, 2], [6.8, 3.2, 5.9, 2.3, 2],
+  [6.7, 3.3, 5.7, 2.5, 2], [6.7, 3.0, 5.2, 2.3, 2], [6.3, 2.5, 5.0, 1.9, 2],
+  [6.5, 3.0, 5.2, 2.0, 2], [6.2, 3.4, 5.4, 2.3, 2], [5.9, 3.0, 5.1, 1.8, 2],
+];
+
+/**
+ * Converts an integer into its one-hot representation and returns
+ * the data as a JS Array.
+ */
+export function flatOneHot(idx) {
```
caisq: So this function is invoked once for every example. Is it possible to make it operate on a batch instead?

iris-fitDataset/data.js
```diff
@@ -0,0 +1,128 @@
...
+  tf.util.shuffle(shuffled);
+  // Split the data into training and testing portions.
+  const numTestExamples = Math.round(IRIS_RAW_DATA.length * testSplit);
+  const numTrainExamples = IRIS_RAW_DATA.length - numTestExamples;
+  const train = shuffled.slice(0, numTrainExamples);
+  const test = shuffled.slice(numTrainExamples);
+  // Split the data into into X & y and apply feature mapping transformations
+  const trainX = tf.data.array(train.map(r => r.slice(0, 4)));
+  const testX = tf.data.array(test.map(r => r.slice(0, 4)));
+  const trainY = tf.data.array(train.map(r => flatOneHot(r[4])));
```
caisq: See my comment above. Might be worth make `flatOneHot` work in a batched manner for efficiency.

iris-fitDataset/python/iris.py
```diff
@@ -0,0 +1,102 @@
+# Copyright 2018 Google LLC. All Rights Reserved.
```
caisq: No need to fork this Python file.

iris-fitDataset/python/iris_data.py
```diff
@@ -0,0 +1,217 @@
+# Copyright 2018 Google LLC. All Rights Reserved.
```
caisq: Similar to above: no need to fork this Python file.

iris-fitDataset/data.js
caisq: Or perhaps even better: is it possible to use `map()`?

iris-fitDataset/data.js
bileschi: It's possible, but then either (1) batching would have to move to inside data.js or (2) the map would have to move outside of data.js.

Kangyi also mentioned (1) in a previous CL.  I opposed early batching in that it's really part of the model, and not part of the data itself, but my opposition is not dogmatic.  I can see how it's an efficiency detail, and so is it's placement.

(2) Moving the mapping function out of data.js would leave data.js simply yielding the shuffled and split dataset.  The preprocoess map call could be added to the .batch call   at line 178, just after the start of iris().

bileschi: Also, we currently perform the transformations before we create a dataset because dataset does not yet have the caching behavior. 

I tried batching and then transforming, as suggested, but it slowed performance by 50%

iris-fitDataset/data.js
bileschi: See above.

iris-fitDataset/python/iris.py
bileschi: done.

iris-fitDataset/python/iris_data.py
bileschi: done

iris-fitDataset/data.js
caisq: OK. But one more comment: `dataSync()` is inefficient. Can you return a `Tensor` from this function, instead of a JavaScript array? I guess that depends on whether `tf.data.array()` works with an array of `Tensor`s.

caisq: FYI, I just tried this and it works:
`const a = tf.data.array([tf.scalar(4), tf.scalar(5), tf.scalar(6)]);`

bileschi: Added a comment here about caching.

PR #204, opened 2019-01-05
Use tf.keras + default regularizers.

mari-linhares:
* Use tf.keras
* Use default regularizer
* Print target sequence in translation training python script
* Change number of epochs from 100 to 20 (Model is overfitting)

Using the default regularizers the model is apparently saved correctly to tensorflow.js, I've tested with the translation example.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/204)


mari-linhares:
Sorry for the wrong description and thanks for fixing it!

PR #209, opened 2019-01-10
[sentiment] Add tfjs-node training script; Fix padding issue

caisq:
- train.js contains the training logic
- train.js uses data.js to load and format the training data
- sequence_utils.js is created to house the truncation/padding logic
  for both the tfjs-node training and in-browser inference
- README.md is updated

### DIFF
--- a/sentiment/.gitignore
+++ b/sentiment/.gitignore
@@ -1,0 +1,4 @@
+*.bin
+*.zip
+model.json
+metadata.json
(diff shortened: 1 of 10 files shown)


caisq:
Gentle ping.

nsthorat:
For the future for the one-liner with a return: 

`localZipFile.on('error', err => reject(err));`

caisq:
@nsthorat Of course... Sorry I was confused for a moment. See https://github.com/tensorflow/tfjs-examples/pull/214

PR #228, opened 2019-02-07
Replace tf.fromPixels with tf.browser.fromPixels

dsmilkov:
- Replace tf.fromPixels with tf.browser.fromPixels
- Make all examples depend on `0.15.1`

tf.browser.fromPixels() works with the latest 0.15.1. This PR is pending on the release of 0.15.1 to update `yarn.lock` files and run examples manually

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/228)

PR #230, opened 2019-02-09
Upgrade tfjs to 0.15.1; Skip examples that dep. on tfjs-node

caisq:
- Also, update the version update script to take into account
  tfjs-node / tfjs-node-gpu in devDependencies
- All example pages are manually tested and confirmed to be working under
  the new version of tfjs.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/230)

### DIFF
--- a/addition-rnn/package.json
+++ b/addition-rnn/package.json
@@ -9,7 +9,7 @@
     "node": ">=8.9.0"
   },
   "dependencies": {
-    "@tensorflow/tfjs": "^0.14.2",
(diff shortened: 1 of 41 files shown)


dsmilkov:
Couple qs:
- Some of these will produce warnings now. Do you want to replace those here, or will you leave specific PRs to handle that (such as https://github.com/tensorflow/tfjs-examples/pull/228)
- Might make sense to wait for 0.15.2 when the last of the changes lands (loadmodel changes) so we don't have to do it twice?

PR #234, opened 2019-02-21
Upgrade non-node examples to tfjs 0.15.3

dsmilkov:
Upgrade non-node examples to tfjs 0.15.3. I tested each upgraded example manually.

Also
- remove usage of deprecated API.
- remove postinstall script which causes problems with `yalc link`, and haven't been of much benefit since we upgrade examples regularly.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/234)

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -47,12 +47,12 @@ for i in $EXAMPLES; do
   echo "building ${EXAMPLE_NAME}..."
   yarn
   yarn build
-  # gsutil mkdir -p gs://tfjs-examples/$EXAMPLE_NAME
-  # gsutil mkdir -p gs://tfjs-examples/$EXAMPLE_NAME/dist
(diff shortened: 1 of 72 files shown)


deploy.sh
```diff
@@ -47,12 +47,12 @@ for i in $EXAMPLES; do
   echo "building ${EXAMPLE_NAME}..."
   yarn
   yarn build
-  gsutil mkdir -p gs://tfjs-examples/$EXAMPLE_NAME
-  gsutil mkdir -p gs://tfjs-examples/$EXAMPLE_NAME/dist
-  gsutil -m cp dist/* gs://tfjs-examples/$EXAMPLE_NAME/dist
+  # gsutil mkdir -p gs://tfjs-examples/$EXAMPLE_NAME
```
caisq: Unintended changes?

dsmilkov: Yes, reverted!

PR #242, opened 2019-03-06
Upgrade browser examples to tfjs v1.0.0

caisq:
Code changes apart from dependency version upgrades:
- In cart-pole: replace `tf.Model` with `tf.LayersModel`
- In data-generator, replace `tf.data.generator()` with `tf.data.func()`. Replace the length-2 array with a "dict" with fields 'xs' and 'ys', as per new tf.data API in v1.0.0.

Fixes https://github.com/tensorflow/tfjs/issues/1328

This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/242)


caisq:
@bileschi Please review the change in data-generator, the breaking API change in tfjs 1.0.0 changed the semantics of `tf.data.generator()`. As a result, the existing code no longer works. There are two ways to fix it:
1. What the current PR is doing, namely to replace `tf.data.generator()` with `tf.data.func()`, which has the same syntax ans semantics as the old `tf.data.generator()`.
2. Change the argument to `tf.data.generator()` into an actual JavaScript generator. 

I opted to do 1, because it's a simpler API in my opinion. Let me now if you are okay with that.

bileschi:
That makes the most sense to me.  Let's also rename the directory from data-generator to data-func in a follow up PR.

PR #245, opened 2019-03-12
[mnist-acgan] Improve command line; add tensorboard and unit tests

caisq:
- Allow easy switching to CUDA GPU training with the `--gpu` flag like in some of the newer examples.
- Add the `--logDir` flag to allow logging the loss values of the generator and the discriminator to tensorboard.
- Add unit tests: gan_test.js

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/245)


caisq:
FYI @nkreeger

PR #254, opened 2019-03-22
[mobilenet] Add timing number without image extraction and preprocessing

caisq:
It turns out that image extraction and preprocessing takes a big fraction of the time.

A typical result (with the new code) looks like:
Done in 76 ms (not including image preprocessing: 21 ms)

This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/254)


caisq:
Thanks, @annxingyuan

PR #261, opened 2019-03-29
[fashion-mnist-vae] Add tensorboard support and unit tests

caisq:
Also in this change:
- Improve the completeness of the doc strings
- Add some unit tests
- Add `tf.tidy()` to `vaeLoss()`
- Make sure the custom ZLayer code uses the correct className static getter and serialization registration method.
- Increase the default number of training epochs to 100 to ensure convergence.

Old generation result:
image

New generation result:
image

This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/261)


caisq:
@tafsiri Following offline discussion, I switched the decoder back to the functional API and added a comment. The latest commit also fixes a wrong logic when tensorboard is used.

PTAL.

PR #279, opened 2019-05-16
Add pretrained resnet50 examples

axinging:
Including v1 and v2.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/279)


caisq:
@axinging Thank you for the pull request! Can you help me understand what the added value of this example provides, given we already have examples that show how to convert a python model into TensorFlow.js format, load it in the browser and use it for inference?

axinging:
Thanks @caisq. 
It seems examples https://github.com/axinging/tfjs-examples/tree/resnet/resnet50v2 is almost the same with the mobilenet, except the model, so I think I should remove this resnet50v2 case.

As to https://github.com/axinging/tfjs-examples/tree/resnet/resnet50, it introduces a different pre-processing model as mentioned (https://github.com/axinging/tfjs-examples/blob/resnet/resnet50/index.js#L67).
I am not sure if this pre-processing has been applied in tfjs-examples. So how do you think?

PR #281, opened 2019-05-30
Make it easy to try and tinker with examples by having an Online IDE (Gitpod)

meysholdt:
hi there 👋!

I'm submitting this PR to answer [a tweet](https://twitter.com/rocksetta/status/1133892484658999296) by @hpssjellis trying Online IDEs for this repository. 

tl;dr: This works just great in [Gitpod](https://www.gitpod.io)!

Having an Online IDE makes it easier for people to try these examples:
* zero setup effort and zero risk of being stuck with a broken setup
* it does not pulute your local file system
* it works in constraint environments: ChromeBooks, iPads, locked-down corporate computers.

This PR contributes a minimal Gitpod configuration. I've only tried this for the `boston-housing` example, but I'm happy to make it work for the other examples, too, if this PR resonates well. 

With this config, Gitpod can:

a) Open and individual example from this repo: 
Try https://gitpod.io/#EXAMPLE=boston-housing/https://github.com/meysholdt/tfjs-examples

b) Open the full repo: 
Try: https://gitpod.io/#https://github.com/meysholdt/tfjs-examples

If you click on (a), the following happens:
1. a docker container is launched for you
2. this repo is checked out via git clone (GitHub OAuth needed).
3. Gitpod runs `cd boston-housing`
4. Gitpod runs `yarn`
5. Gitpod runs `yarn watch`
6. When `yarn warch` is *ready*, you can click on "open preview" in a popup message from Gitpod.
7. What you should see is the following:

image

Disclosure: I work on Gitpod.

What we see in the `command` section in `.gitpod.yml` is a just `bash` code and can easily be extended/changed to do what ever is great for this repo. 

Since this repo contains many examples, I'm proposing to use Gitpod's [support for env variables](https://www.gitpod.io/docs/47_Environment_Variables/) to pass in a variable (e.g. `EXAMPLE=boston-housing`) which can then be accessed in Gitpod's config/container. Said `bash` code uses this variable to either open a specific example or print general usages messages.

There is one known issue here: In step (6), Gitpod could automatically open the preview when `yarn watch` opens the port, but it seems that the port opens *much* earlier than `yarn watch` actually serves content, causing Gitpod's preview to timeout. I'm happy to help fix that. 

let me know what you think.


meysholdt:
I signed it!

meysholdt:
in my original description I also wrote:
> There is one known issue here: In step (6), Gitpod could automatically open the preview when `yarn watch` opens the port, but it seems that the port opens *much* earlier than `yarn watch` actually serves content, causing Gitpod's preview to timeout. I'm happy to help fix that. 

This is fixed now by introducing a browser handler that delegates to Gitpod's Preview.

tafsiri:
Thanks for this! We'll take a closer look and reply more fully as soon as we can. However I did want to ask, is there a way for a user to set this up directly in gitpod without the .gitpod files being in the repo, (something like project settings or templates), I realize this may not be one-click, but I'm curious to know if that is an option.

meysholdt:
Hi Yannick! Thank you for your response and for looking at this. Let me know if any questions come up.

I assume you want to keep the config files in this repo as few as possible. I'm happy to help with that. I can remove `.gitpod-preview.sh` from this PR once we have a fix for https://github.com/gitpod-io/gitpod/issues/596. 

Keeping the `.gitpod.yml` automates the steps that you have documented in your `README.md` files with a few lines of `bash` code in `yaml`. No magic here. I believe that's nice for users because they spend less time figuring out the setup and more time with the actual examples. 

The links in the README to open examples in the online IDE I believe can increase the number of users that actually try the examples. With local development you need to worry about your setup, changes to your local setup breaking other projects you work on, leaving behind leftover files and setup that causes trouble down the road...At least for me, this was a reason to NOT try many examples in the past, even though I was curious. In contrast, an online IDE is a risk-free and quick way to try an example. The README files are excellent locations to show users this possibility.

Keeping the `.gitpod.yml` also allows to run `yarn install` etc. *before* users open workspaces. So users don't need to wait for `yarn install` to finish. This can be enabled by configuring the [Gitpod App](https://github.com/apps/gitpod-io/) on the repo. I see that you already have the Reviewable app configured. The Gitpod app will execute the commands from the `.gitpod.yml` on `git push`, just like a CI/CD build would do. The difference is that here we don't get artifacts for prod, we get pre-built workspaces. I think this option can be interesting if you want to have *all* examples prebuilt. For opening individual examples I think it is not critical here because individual examples compile reasonably quickly. The app is completely optional and only needed to have pre-built workspaces.

To try the experience without `.gitpod.yml`, you can click on this link *before* merging this PR:
https://gitpod.io/#https://github.com/tensorflow/tfjs-examples

For comparison, this is the experience for a chosen example with `.gitpod.yaml`:
https://gitpod.io/#EXAMPLE=boston-housing/https://github.com/meysholdt/tfjs-examples

Let me know what you think :)

tafsiri:
@meysholdt Thanks for the detailed info! We will definitely take a look. I'll give you a bit of a warning that it might take time (longer than you expect) to get this decided internally. This is both to give team members a chance to experiment with it, and seeing if we need any approvals outside of our dev team, e.g. for things like the adding the github app on the repo or any license review stuff.

tafsiri:
@meysholdt One more question. Are these workspaces shared or does each user get their own vm/container so that changes they make are isolated?

meysholdt:
Hi Yannick! 

I'll be patient :) Thank you for the explanation and I'm happy to hear your team members will look at it, too. Please extend my offer to them to contact me, in case questions come up.

Users get their own containers. 

In fact, a user gets a new container each time she/he clicks on a `https://gitpod.io/#https://github.com/...` link or on a [snapshot](https://twitter.com/rocksetta/status/1134519957989912576). Users can have up to eight containers (meaning: eight workspaces) running simultaneously. Users can stop workspaces if they choose to. After 30min of inactivity, workspaces stop automatically. The number of stopped workspaces a user can have is not limited. Stopped workspaces can be started again, and all the old files will be there, even the ones that have not been git-committed/pushed. A user can see her/his list of running/stopped workspaces here: https://gitpod.io/workspaces/.

Users have several options to share...they can
* create snapshots from their workspaces and share the snapshot URL: Other users get their own copy of the workspace, without being able to modify the original. 
* live-share the with other users: Multiple users work in the same container.
* expose web applications running in the workspace to other users by sharing the URL.

PR #286, opened 2019-07-02
factor out dist dir

bileschi:
Feel like this is not the right way to do this.  PTAL

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/286)


nsthorat:
QQ -- where are these images actually used? I don't see them referred to in the code.

bileschi:
They are the extension icons.  They are referenced in manifest.json

bileschi:
Should this use rimraf and some npm version of the other file util commands?

PR #288, opened 2019-07-03
[electron] Add electron example

caisq:
Example screenshot: 
screenshot-2

This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/288)


electron/README.md
```diff
@@ -0,0 +1,113 @@
...
+## How to use this example
+
+1. Build and launch the app using:
+
+   ```sh
+   yarn
+   yarn start
+   ```
+
+   If you have a CUDA-enabled GPU on your machine and have set up
```
nkreeger: I'd really consider leaving this out - I know this is the most performant option - but shipping desktop applications that require system CUDA is something unreasonable for users. Those of us working on models might have CUDA setup - but most users will not.

Imagine this scenario:
"In order to use my app - please only use Linux or Windows and painfully install this very specific NVIDIA CUDA driver".

When `tfjs-headless-nodegl` launches - this is an option for GPU acceleration w/o expecting users to have a non-default driver.

electron/package.json
```diff
@@ -0,0 +1,39 @@
...
+  },
+  "main": "main.js",
+  "scripts": {
+    "start": "electron ."
+  },
+  "build": {
+    "linux": {
+      "target": [
+        {
+          "target": "deb",
```
nkreeger: Are we only supporting linux for now?

electron/README.md
nsthorat: +1

PR #322, opened 2019-07-28
Add webworker example

WenheLI:
This PR adds an example that trains an rnn network under webworker. In this way, the whole training process will not block UI.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/322)


WenheLI:
@dsmilkov - Hi Daniel, can you help look at this failure? 
It fails at `lstm-text-generation` test. I did not make any commits related to it. It should be a problem of testing env, from my perspective.

PR #323, opened 2019-08-12
[snake-dqn] Fix a bug in copyWeights(); Upgrade to tfjs 1.2.7

caisq:
Workaround for https://github.com/tensorflow/tfjs/issues/1807

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/323)


snake-dqn/dqn.js
```diff
@@ -68,5 +68,22 @@ export function createDeepQNetwork(h, w, numActions) {
...
+    originalDestNetworkTrainable = destNetwork.trainable;
+    destNetwork.trainable = srcNetwork.trainable;
+  }
+
   destNetwork.setWeights(srcNetwork.getWeights());
+
+  // Weight orders are inconsistent when the trainable attribute doesn't
+  // match between two `LayersModel`s. The following is a workaround.
+  // TODO(cais): Remove the workaround once the underlying issue is fixed.
+  if (originalDestNetworkTrainable != null) {
```
bileschi: This check is weird.  If the destNetwork.trainable == null, and the srcNetwork.trainable ==, say, True, then this function does nothing?

bileschi:
___
*[snake-dqn/dqn.js, line 86 at r1](https://reviewable.io/reviews/tensorflow/tfjs-examples/323#-Lm2msqlhCJ3n0xAmIiW-r1-86:-Lmf0O9w57AXrBlaxeKc:ban0eaj) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/2c4f93106912ac1b32895c9675f974fb7ff30850/snake-dqn/dqn.js#L86)):*

The `trainable` property of a `LayersModel` is either True or False; it can't be `null`.

See
https://github.com/tensorflow/tfjs/blob/master/tfjs-layers/src/engine/container.ts#L509

The `originalDestNetworkTrainable` is `null` if and only if the `trainable` properties of the two `LayersModel` instances are the same, in which case it makes sense to do nothing.
</blockquote></details>

Ok, that makes sense, can you add that last sentence as a comment above this simple check?

PR #331, opened 2019-08-29
Fix custom layer bug

bileschi:
Wait for document.readyState == 'complete' before inspecting image contents.

Also moves babel configuration out of package.json and into a .babelrc file, to match the other examples.


bileschi:
fixes https://github.com/tensorflow/tfjs/issues/1911

nsthorat:
I'm going to fix this upstream in core

PR #336, opened 2019-10-20
Closure not needed for `let` variable

mehulmpt:
`idx` is not required. `i` is already in a closure as `let` is used.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/336)


mehulmpt:
@googlebot I signed it!

PR #344, opened 2020-02-29
Update index.js

hacker-joe:


mihaimaruseac (changes requested):
I don't think this makes sense in prod

PR #348, opened 2020-05-29
Update a few examples to 2.0

tafsiri:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/348)


tafsiri:
Generally we should update examples any time we have opportunity to touch them. This starts that process.

tafsiri:
Thanks Na!

PR #352, opened 2020-08-22
loading tfjs-node-gpu is handled by the --gpu flag

josephrocca:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/352)


mattsoulanille (approved):
LGTM. Thanks!

Other reviewers [see these lines of train.js](https://github.com/tensorflow/tfjs-examples/blob/master/fashion-mnist-vae/train.js#L186-L192) for how this works.

PR #353, opened 2020-09-13
Fix learningRate param in snake-dqn

pskucherov:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/353)


caisq (approved):
Thanks for the fix, @pskucherov

PR #354, opened 2020-09-18
Update index.js

coby1kenobi:
I'm not really sure about this one since I'm pretty new to this whole thing with tensorflow.js and coding in general. const data[] is defined twice and used in the later function it is defined in. I would really appreciate knowing why it is also defined there if that's really the case.

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/354)


coby1kenobi:
@googlebot I signed it!

pyu10055 (approved):
Thanks

PR #361, opened 2020-10-08
[Interactive Visualizer] Add logic to fetch model, metadata, test data, and display test images

cbeausei:
Screenshot with metadata from https://storage.googleapis.com/tfhub-visualizers/google/aiy/vision/classifier/plants_V1/1/metadata.json:

5ioBWzdhLLJCvPL

Also adds unit tests (tested locally).

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/361)


lina128 (commented):
Hi Clement, overall looks good to me. I just have a high level question about whether we should use es2015 or 2017. See my comments below. Thank you! 

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/361#-:-MJ8LMPPE_e2bRWdDSCV:b-onlpqg)* status: 0 of 1 approvals obtained (waiting on @cbeausei, @lina128, and @pyu10055)

___
*[interactive-visualizers/src/app/app.component.ts, line 63 at r1](https://reviewable.io/reviews/tensorflow/tfjs-examples/361#-MJ8HFxFCklNttA-9Ku3:-MJ8HFxFCklNttA-9Ku4:b4fvkct) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/5d82fb42132b588a2a89c952afbe788cdb04605c/interactive-visualizers/src/app/app.component.ts#L63)):*
> ```
> https://github.com/angular/angular/issues/31730
> ```
In this case, shall we change to es2015? The reason we want to go with es2017 is to have async/await natively. If this doesn't work with Zone.js, it forfeits the purpose. Do you see other benefit using es2017? 
___
*[interactive-visualizers/src/app/app.component.ts, line 110 at r1](https://reviewable.io/reviews/tensorflow/tfjs-examples/361#-MJ8G-o538czAgylts-T:-MJ8G-o538czAgylts-U:bt170zl) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/5d82fb42132b588a2a89c952afbe788cdb04605c/interactive-visualizers/src/app/app.component.ts#L110)):*
> ```XML
>   }
> 
>   getAssetsUrlPrefix(): string {
> ```

Can we add an annotation about what the modelMetadataUrl looks like? It will help understand the code in this block.
___
*[interactive-visualizers/src/app/app.component.ts, line 112 at r1](https://reviewable.io/reviews/tensorflow/tfjs-examples/361#-MJ8Eynx2qck1ZLjJm7c:-MJ8Eynx2qck1ZLjJm7d:b9t7ha6) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/5d82fb42132b588a2a89c952afbe788cdb04605c/interactive-visualizers/src/app/app.component.ts#L112)):*
> ```XML
>   getAssetsUrlPrefix(): string {
>     const metadataFileName = this.modelMetadataUrl.split(
>         '/')[this.modelMetadataUrl.split('/').length - 1];
> ```

It looks like split is done twice here, can we do:
`const metadataFileName = this.modelMetadataUrl.split('/').pop();`
___
*[interactive-visualizers/src/app/app.component.ts, line 118 at r1](https://reviewable.io/reviews/tensorflow/tfjs-examples/361#-MJ8GHde5_qUM5lUpLNz:-MJ8GHde5_qUM5lUpLO-:b2nreh6) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/5d82fb42132b588a2a89c952afbe788cdb04605c/interactive-visualizers/src/app/app.component.ts#L118)):*
> ```XML
>   getTestImagesUrlPrefix(): string {
>     const imageIndexFileName = this.testImagesIndexUrl.split(
>         '/')[this.testImagesIndexUrl.split('/').length - 1];
> ```

Same as above.

cbeausei:
Sounds indeed like we should rather use es2015, updated. Also addressed other comments. Thanks for fast review!

lina128 (approved):
LGTM! There's some Angular errors in the test, please fix.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/361#-:-MJ8hyHz7AczIfxxM9kC:b-c1a8ry)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @cbeausei and @pyu10055)

cbeausei:
Thanks! Fixed the error.

lina128 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/361)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @cbeausei and @pyu10055)

___
*[interactive-visualizers/src/app/app.component.ts, line 20 at r3](https://reviewable.io/reviews/tensorflow/tfjs-examples/361#-MJ8r9Zd6KqIY1qi51bK:-MJ8r9ZeBtkCrx5FdK-r:bbv5vz0) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/94b12c400c3e9fe02143c2dc2b8ba9bf06feb9bc/interactive-visualizers/src/app/app.component.ts#L20)):*
> ```XML
> import {Component} from '@angular/core';
> import * as tf from '@tensorflow/tfjs';
> tf.setBackend('cpu');
> ```

await tf.setBackend('cpu'), this probably needs to be put in ngOnInit

lina128 (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/361)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @cbeausei and @pyu10055)

___
*[interactive-visualizers/src/app/app.component.ts, line 20 at r3](https://reviewable.io/reviews/tensorflow/tfjs-examples/361#-MJ8r9Zd6KqIY1qi51bK:-MJ8rXHs866bbXTXexqu:blb5v1y) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/94b12c400c3e9fe02143c2dc2b8ba9bf06feb9bc/interactive-visualizers/src/app/app.component.ts#L20)):*
<details><summary><i>Previously, lina128 (Na Li) wrote…</i></summary><blockquote>

await tf.setBackend('cpu'), this probably needs to be put in ngOnInit
</blockquote></details>

Also should we try webgl first, if fail, fall back to cpu. I think the package automatically does this. Why do you need to pin to cpu backend? It may be too slow to run some image models.

cbeausei:
I'm a bit confused as it seems this is the one warning showing when running the test. If this is the default behavior, indeed it sounds better not to add this to the main TS file.

I'm adding it therefore to the test file, WDYT?

cbeausei:
Oh you were mentioning the linter issues, I just realized that, will address!

cbeausei:
Addressed linter errors.

lina128 (approved):
Hi Clement, LGTM overall. I just left a nit comment.

> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/361#-:-MJ9TqMD18Us2WhgOhZ2:b-jfzciw)* status: :shipit: complete! 1 of 1 approvals obtained (waiting on @cbeausei, @lina128, and @pyu10055)

___
*[interactive-visualizers/src/app/app.component.ts, line 55 at r5](https://reviewable.io/reviews/tensorflow/tfjs-examples/361#-MJ9TMah7BJov0hCwMn8:-MJ9TMah7BJov0hCwMn9:b1ne0on) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/59bfaca0a20142bec25dd4c36a7a45f4946819b5/interactive-visualizers/src/app/app.component.ts#L55)):*
> ```XML
>       throw new Error(NO_MODEL_METADATA_ERROR_MESSAGE);
>     }
>     const modelMetadataUrl = urlParams.get('modelMetadataUrl') || '';
> ```

Above assertion already prevent falsey cases, so I guess you don't need `|| ''`?

cbeausei:
True indeed, thanks for the review Na!

PR #372, opened 2020-10-23
remove testing dependency on jasmine_util and update dependencies

tafsiri:
This removes dist imports from tjfs-core.

Updates the following examples to tfjs 2.6.0:
abalone-node, date-conversion-attention, fashion-mnist-vae, jena-weather, lstm-text-generation, mnist-acgan, snake-dqn

jena-weather also had a code update to match the current tfjs-data api.
Updating sentiment turned up a bug that may be an issue in core. So I'll look at that separately and update once there is a diagnosis/fix.

This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/372)

### DIFF
--- a/abalone-node/package.json
+++ b/abalone-node/package.json
@@ -9,7 +9,7 @@
     "test": "node run_tests.js"
   },
   "dependencies": {
-    "@tensorflow/tfjs-node": "^1.3.2",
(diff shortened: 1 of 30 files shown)


caisq (approved):
Thanks a lot, @tafsiri !

PR #385, opened 2021-01-07
update mnist-core, sentiment and, date-conversion-attention to tfjs 3.0

tafsiri:
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/385)


tafsiri:
cc @caisq @bileschi There is an RC of the new 3.x available if you want to check compatibility in advance of the final release. Details [available here](https://github.com/tensorflow/tfjs/releases/tag/tfjs-v3.0.0-rc.0).

caisq:
@tafsiri Thanks for opening this PR. 

What action are you expecting from me and @bileschi about this PR? I've just manually checked the date-conversion-attention example and it works fine under 3.0.0-rc.0. Testing all examples may take longer. But I don't see why testing the examples should block the final release, as the changes aren't that dramatic (there may be a small number of places where we use *Strict method and those need to change).

tafsiri:
@caisq Sorry I should have made that clearer, no action on this PR at this point, and it won't block the release at all. Once 3.0 final is out i'll update this PR and mark it ready for actual review/merge

It was more of a heads up in case you wanted a headstart on checking the other examples. We are also planning a fixit week next week to do similar upgrades for tfjs-models (which will involve more changes than I expect the examples with) and stage those updates in master. But its less urgent for the examples than it is for the models. So just an FYI so you can plan ahead.

tafsiri:
I made this as a pr purely to make sure it passes CI.

tafsiri:
@caisq @bileschi this is now ready for review and updates the examples to tfjs 3.0.0

bileschi (approved):

caisq (approved):

PR #673, opened 2022-01-20
[chrome-extension] Update the TFJS chrome extension example to v3 and to use mobilenet model package directly.

jinjingforever:
- Rename background.js to service_worker.js to make things more obvious (please compare it with the now deleted background.js for diffs. Github didn't recognize the rename)
- Move the image processing logic to content script (service worker cannot access DOM)
- Use the mobilenet model directly and remove unnecessary code.
- Add more info to README

Tested locally and it works as before.

Fixes https://github.com/tensorflow/tfjs/issues/5889

---
This change is [Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/673)


pyu10055 (changes requested):
> Reviewed 8 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/673)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @jinjingforever)

___
*[chrome-extension/README.md, line 50 at r1](https://reviewable.io/reviews/tensorflow/tfjs-examples/673#-Mtt0ezw5teDXbFSQ8w8:-Mtt0ezw5teDXbFSQ8w9:b2bgz03) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/7a4b81656ad0e97ed3624b7a057a1e6947f2ff42/chrome-extension/README.md#L50)):*
> ```Markdown
> canvas, and sends the data back to the service worker.
> - After the service worker receives the image data, it runs the mobilenet model
> with the data, and gets the prediction results. It then sends to results back
> ```

does the service worker reload the model on every request? If so, will the model files be cached by the browser?
___
*[chrome-extension/src/content.js, line 30 at r1](https://reviewable.io/reviews/tensorflow/tfjs-examples/673#-Mtt13Of3KFuQ1-woEfm:-Mtt13Of3KFuQ1-woEfn:bn2ee3l) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/7a4b81656ad0e97ed3624b7a057a1e6947f2ff42/chrome-extension/src/content.js#L30)):*
> ```JavaScript
> // The minimum image size to consider classifying.  Below this limit the
> // extension will refuse to classify the image.
> const MIN_IMG_SIZE = 128;
> ```

why there needs to have a minimum size? Is required by the data transfer mechanism or for the benefit of more accurate prediction?
___
*[chrome-extension/src/service_worker.js, line 3 at r1](https://reviewable.io/reviews/tensorflow/tfjs-examples/673#-Mtsy3viDy0Df5fqhAf4:-Mtsy3viDy0Df5fqhAf5:bb6ht37) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/7a4b81656ad0e97ed3624b7a057a1e6947f2ff42/chrome-extension/src/service_worker.js#L3)):*
> ```JavaScript
> /**
>  * @license
>  * Copyright 2019 Google LLC. All Rights Reserved.
> ```

2022

jinjingforever (commented):
> *[Reviewable](https://reviewable.io/reviews/tensorflow/tfjs-examples/673)* status: 1 change requests, 0 of 1 approvals obtained (waiting on @pyu10055)

___
*[chrome-extension/README.md, line 50 at r1](https://reviewable.io/reviews/tensorflow/tfjs-examples/673#-Mtt0ezw5teDXbFSQ8w8:-MttWrE98qST1ze-Mazu:b-39mdqy) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/7a4b81656ad0e97ed3624b7a057a1e6947f2ff42/chrome-extension/README.md#L50)):*
<details><summary><i>Previously, pyu10055 (Ping Yu) wrote…</i></summary><blockquote>

does the service worker reload the model on every request? If so, will the model files be cached by the browser?
</blockquote></details>

I think the lifecycle of service worker is managed by Chrome. It might get unloaded at any time (for example, after x minutes of idle). When it is loaded again, the model will be reloaded, but it will be from the cache which should be fast. 

I didn't see it getting unloaded from my (limited) testing, but it should work correctly even it does.
___
*[chrome-extension/src/content.js, line 30 at r1](https://reviewable.io/reviews/tensorflow/tfjs-examples/673#-Mtt13Of3KFuQ1-woEfm:-MttXq2b7A7XhJiMRzXt:b-62c25d) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/7a4b81656ad0e97ed3624b7a057a1e6947f2ff42/chrome-extension/src/content.js#L30)):*
<details><summary><i>Previously, pyu10055 (Ping Yu) wrote…</i></summary><blockquote>

why there needs to have a minimum size? Is required by the data transfer mechanism or for the benefit of more accurate prediction?
</blockquote></details>

Yeah, I didn't touch this part. I think the accuracy is one reason. The other reason is that the extension would overlay the results on top of the image so it cannot be too small. 
___
*[chrome-extension/src/service_worker.js, line 3 at r1](https://reviewable.io/reviews/tensorflow/tfjs-examples/673#-Mtsy3viDy0Df5fqhAf4:-MttYIvvAof3rhtDxFPg:b-896fix) ([raw file](https://github.com/tensorflow/tfjs-examples/blob/7a4b81656ad0e97ed3624b7a057a1e6947f2ff42/chrome-extension/src/service_worker.js#L3)):*
<details><summary><i>Previously, pyu10055 (Ping Yu) wrote…</i></summary><blockquote>

2022
</blockquote></details>

Done.

pyu10055 (approved):

nelson-liu:
Thanks for updating this example, it's very helpful! I had a few questions with respect to how to best use service workers with TensorflowJS models.

(1) In my experience, it seems like the model can get unloaded pretty frequently. Is there a better way to make the model stick around, or is this just a fundamental limitation that manifest V3 imposes?
(2) This example considers a setting where the model is largely receiving one example for inference at a time. What if the user triggers inference for many messages at once, causing the model to receive many messages simultaneously? The model would be stuck performing inference on many examples, which might be pretty slow---ideally, you'd want some sort of queue from which a model could take 1 or `batch_size` examples, perform inference, and then send back results. However, this seems a bit tough to do with manifest v3 because the service worker can be killed at any time---any thoughts or tips on what a reasonable thing to do would be?

Thanks in advance!

PR #1067, opened 2023-04-10
fix tsne-mnist-canvas

alicialics:
The current tsne-mnist-canvas example has a dependency on @tensorflow/tfjs@0.14.1 so it does not work right now:
https://storage.googleapis.com/tfjs-examples/tsne-mnist-canvas/dist/index.html

The drops back the @tensorflow/tfjs to 0.14.1 to fix API incompatibility.

@tensorflow/tfjs-tsne is archived unfortunately.  I am able to update @tensorflow/tfjs-tsne to be compatible with latest @tensorflow/tfjs@4.4.0 however the source repo needs to be unarchived and new versions be published into npm.

### DIFF
--- a/tsne-mnist-canvas/package.json
+++ b/tsne-mnist-canvas/package.json
@@ -5,7 +5,7 @@
   "license": "Apache-2.0",
   "private": true,
   "dependencies": {
-    "@tensorflow/tfjs": "^1.3.2",
+    "@tensorflow/tfjs": "0.14.1",
     "@tensorflow/tfjs-tsne": "0.2.0",
     "d3": "^5.2.0"
   },
--- a/tsne-mnist-canvas/yarn.lock
+++ b/tsne-mnist-canvas/yarn.lock
@@ -824,60 +824,126 @@
     chrome-trace-event "^1.0.2"
     nullthrows "^1.1.1"
 
+"@protobufjs/aspromise@^1.1.1", "@protobufjs/aspromise@^1.1.2":
+  version "1.1.2"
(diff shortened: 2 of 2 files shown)


romanzenka:
I would very much like this work to make it in, as I'd like to use the demo. What is the status?

alicialics:
> I would very much like this work to make it in, as I'd like to use the demo. What is the status?

@romanzenka just pin @tensorflow/tfjs to 0.14.1 and you should be good to go.

romanzenka:
> @romanzenka just pin @tensorflow/tfjs to 0.14.1 and you should be good to go.

It does not actually work. I think the new tensorflow has extracted a module for webgl, so while I get a little bit further, the demo still would not run. Will keep investigating.

alicialics:
> > @romanzenka just pin @tensorflow/tfjs to 0.14.1 and you should be good to go.
> 
> It does not actually work. I think the new tensorflow has extracted a module for webgl, so while I get a little bit further, the demo still would not run. Will keep investigating.

Do you mind sharing repo instructions and your tfjs version?  Happy to help but need a bit more information.

romanzenka:
> Do you mind sharing repo instructions and your tfjs version? Happy to help but need a bit more information.

Sure: I am on commit 6d5986a6c73f62753f2253e263214ac02706445f of  tsne-mnist-canvas. 

I did this to pin tfjs:

`"@tensorflow/tfjs": "0.14.1",``

Then I did `yarn watch` and to my great surprise, it worked! The only explanation I have (since I have not changed anything since previous time), is that there must have been a glitch in my Chrome version that got fixed somehow? Thank you for your help!

PR #1120, opened 2023-09-03
Corrected cart pole angle

GregorAltstaedt:
Corrected the angle at which the cart pole leans (it was reversed)


GregorAltstaedt:
This might be the smallest pull request. Is there someone who can review it? :)

PR #1399, opened 2025-04-03
Fix Cart-Pole Render Angle

Kinvert:
When rendering Cart Pole, the pole is at the wrong angle. This one liner fixes it.

I saw you want issues created. I didn't see how to do it here so I followed the link you provided in your README.md and it went to another repo. I created the issue here: https://github.com/tensorflow/tfjs/issues/8538

My son and I noticed it when I was trying your demo. I dug through the code and found this, and the sign swap works.


Kinvert:
For some reason your CLA thing is tying to other accounts instead of my GitHub account. I tried opening in a private window and it wants me to sign in with my Google account. That has a different email address. How do I sign this CLA without mixing up email addresses?

